Issue #30, opened 2025-09-20
Include integration tests in code-coverate results


rstata:
Many tests of the Tenant API implementation (e.g., code found in `tenant/runtime`) are performed by integration tests that depend on code generation.  The source for these tests is found in `integrationTest` source sets (e.g., `tenant/runtime/integrationTest`), but are built an run by the integration-test Gradle projects (e.g., `tenant/runtime-integration-tests`).  The off-the-shelf configuration of Jacaco does not include those tests in its coverage reports, so those reports are not as useful as they could be.

The goal of this issue is to configure Jacoco so that it includes integration tests as testing code from their respective main source sets.

PR #31, opened 2025-09-20
Include integration test in Jacoco coverage reports

rstata:
## Description

See Issue #30.  Copying the description: "Many tests of the Tenant API implementation are performed by integration tests that depend on code generation. The source for these tests is found in `integrationTest` source sets, but are built an run by the integration-test Gradle projects (e.g., `tenant/runtime-integration-tests`). The off-the-shelf configuration of Jacaco does not include those tests in its coverage reports, so those reports are not as useful as they could be. The goal of this issue is to configure Jacoco so that it includes integration tests as testing code from their respective main source sets."

## How has this been tested?
Manual creation and inspection of coverage reports.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ x] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.
- [ x] I have added tests to cover my changes.
- [x ] All new and existing tests passed.


settings.gradle.kts
```diff
@@ -15,8 +15,37 @@ plugins {
 
 rootProject.name = "viaduct"
 
-includeBuild(".")
-includeBuild("included-builds/core")
+includeBuild(".") // TODO - is this needed?
+includeBuild("included-builds/core") {
+    dependencySubstitution {
```
skevy, line 20: I'm very surprised this is necessary. These are all the _default_ dependency substitutions.

settings.gradle.kts
```diff
@@ -15,8 +15,37 @@ plugins {
 
 rootProject.name = "viaduct"
 
-includeBuild(".")
-includeBuild("included-builds/core")
+includeBuild(".") // TODO - is this needed?
```
skevy, line 18: Yes, it's necessary. It allows the `viaduct-bom` project to be resolved by other included builds (like the demo apps).

settings.gradle.kts
rstata: i was surprised to - wasted a bunch of time on this actually - but it seems like if you have one substitution in a project you need to have all substitutions in the project - and if you don't you get strange failures.

skevy: 👍

build-logic/src/main/kotlin/jacoco-integration-tests.gradle.kts
```diff
@@ -0,0 +1,142 @@
+/**
+ * Add this to integration-test projects (e.g., tenant/runtime-integration-tests)
+ * 
+ * It will generate an exec file for the integration tests and then combine that
+ * with the unit tests from the base project to generate a full coverage report
+ */
+
+import org.gradle.api.attributes.*
```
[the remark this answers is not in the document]
rstata: i'm not sure i see that we're using quasi-internal details here.  that said, i did take a pretty serious look at an alternative.  in particular, the jacoco-report-aggregation plugin.  i think there are things that i'm doing in the tasks created by `registerCoverageTask` that might be built into jacoco report aggregation.  that said, any complexity here is due to the fact that the base project and the integration-test project are in different builds in our composite build.  this means we should be using configurations to connect these tasks.  in attempting a quick prototype that used jacoco report aggregation, i was quickly recreating pretty much everything in this PR.

PR #44, opened 2025-09-26
Sends Slack alerts when errors occur in CircleCI jobs.

gperalta-ductworks:
Sends Slack alerts when errors occur in CircleCI jobs.

## Description

Ensure teams are notified in real time about failures in CircleCI pipelines. This reduces response time, helps identify issues quickly, and improves pipeline reliability.
Adds Slack alert integration that posts error notifications from CircleCI pipelines directly to configured Slack channels.

Verified by running failing jobs in CircleCI and confirming that Slack alerts were triggered with the correct error details in the target channel. Successful runs were also checked to ensure no false alerts.

Documentation: A minimal README with environment variable setup (CIRCLE_TOKEN, SLACK_ACCESS_TOKEN, SLACK_DEFAULT_CHANNEL) and instructions for enabling the Slack integration in

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)


Goncalo-Almeida (approved):

.circleci/config.yml
fireboy1919, line 1: Two comments:

1.  It looks like there are some fail steps that have been removed without being replaced with slack messages.
2. It looks like you're not passing along the actual message describing what failed.  Looks like that's supported in by the slack orb, so you probably should.

.circleci/config.yml
```diff
@@ -41,31 +42,21 @@ commands:
...
-            echo "💡 Troubleshooting tips:"
-            echo "  1. Check the logs above for specific error messages"
-            echo "  2. Verify all dependencies are properly installed"
-            echo "  3. Review recent changes that might have caused this failure"
-            exit 1
+          name: Send alerts to slack when something fails
           when: on_fail
+          #Clone the repo to be sure that script circleci_pipeline_metadata.sh is always available.
+          command: |
+            git clone --depth=1 "https://github.com/gperalta-ductworks/viaduct.git" script_metadata
```
fireboy1919, line 55: Instead of doing this (which is pretty expensive), you can add the script to the CircleCI workspace during the first step of checkout.  Then it will always be available.

PR #235, opened 2025-12-02
docs: Break out release runbook and update it

rstata:
## Description

The `RUNBOOK.md` file was getting very long exclusively because of the instructions for cutting a release.  This PR breaks those out into its own file.  It also adds corrections and clarifications to the release instructions.

Docs only

### DIFF
--- a/RELEASE-RUNBOOK.md
+++ b/RELEASE-RUNBOOK.md
@@ -324,6 +324,8 @@ Once the artifacts are published, we need to update the standalone copies of the
   - `cli-starter` → `viaduct-graphql/cli-starter`
   - `ktor-starter` → `viaduct-graphql/ktor-starter`
 
+**It's important to do this on the release branch!**
+
 We do this with a copybara script. For each demoapp run:
 
 ```bash


RELEASE-RUNBOOK.md
```diff
@@ -0,0 +1,394 @@
...
+  - Pushes a `vX.Y.Z` tag to Github.
+  - Create a draft Github release with changelog.
+
+### 9) Verify publications
+
+Log in to Sonatype and the Gradle Plugin Portal to verify the artifacts are live (credentials in shared 1Password vault).
+
+TODO: A bit more detail on this step...
+
+### 10) Publish and verify standalone apps
```
fireboy1919: The description here does not guarantee that the version that is published is the release version.  That is why I recommend using the script we hvae for this; it will double check that you're publishing a release version.

If we don't want to do that, it's fine, but we should probably mention in this step that you need to be in a release branch and the release version has to be committed for it to run.

rstata: i added this comment.

PR #239, opened 2025-12-05
feat(service-wiring): add GraphiQL HTML generator with Viaduct customizations

fireboy1919:
**PUBLIC**
feat(service-wiring): add GraphiQL HTML generator with Viaduct customizations

Add GraphiQL IDE resources to service/wiring module with custom plugins and patches for GraphQL Java compatibility.

- Add GraphiQLHtmlCustomizer in build-logic to download and customize official GraphiQL CDN example
- Add generateGraphiQLHtml Gradle task for manual regeneration
- Add introspection-patch.js to fix GraphQL Java compatibility (GraphQL Java omits empty arrays that GraphiQL 5 requires)
- Add global-id-plugin.jsx for encoding/decoding Viaduct Global IDs
- Add jsx-loader.js for runtime JSX transpilation
- Add GraphiQLHtml.kt helper to read HTML from resources

## Test Plan
- [x] Verified `./gradlew :core:service:service-wiring:generateGraphiQLHtml` generates the customized HTML
- [x] Verified `./gradlew :core:service:service-wiring:compileKotlin` builds successfully
- [x] Verified generated HTML preserves original GraphQL Contributors copyright
- [x] Verified Viaduct disclaimer is included explaining customizations


demoapps/starwars/src/main/kotlin/com/example/starwars/service/graphiql/StaticJsController.kt
```diff
@@ -9,17 +9,15 @@ import io.micronaut.http.annotation.Produces
 class StaticJsController {
     @Get("/jsx-loader.js")
     @Produces("application/javascript")
-    fun jsxLoader(): HttpResponse<String> {
-        val resource = this::class.java.classLoader.getResource("graphiql/js/jsx-loader.js")
-        val content = resource?.readText() ?: return HttpResponse.notFound()
-        return HttpResponse.ok(content).contentType("application/javascript")
-    }
+    fun jsxLoader(): HttpResponse<String> = serveJs("jsx-loader.js")
```
rstata: why do you have the function `graphiQlHtml()` for the page and then have this slightly different `serveJs` API here?  do we really need an abstraction here?  what was wrong in both cases with the `HttpResponse<String>` invocation that you're replacing?

PR #303, opened 2026-03-11
feat(ci): restructure build-and-test workflow to reduce CI latency

rstata:
The previous workflow had a build matrix followed by a test matrix, with
tests unable to start until every build coordinate finished (GitHub Actions
`needs:` is all-or-nothing across a matrix).  Linting and coverage
verification were also sequenced behind the full test matrix.

The new structure shortens the critical path by:

- validate-inputs: validates workflow dispatch inputs, runs the Python CI
  script unit tests, and computes the deep coordinate and wide matrix for
  downstream jobs.  CI script tests run here because validate-inputs
  executes those scripts directly — a script regression would produce
  untrustworthy outputs, so testing them first halts everything before any
  downstream job acts on bad data.

- build-deep: a single coordinate (first OS × first Java from the input
  lists, defaulting to ubuntu-latest × 17).  Runs assembly and uploads a
  build artifact for downstream jobs to consume.

- test-deep: same single coordinate, depends on build-deep.  Runs tests,
  coverage report generation, and coverage threshold verification via the
  new `testWithCoverage` Gradle task.

- docs-verify: same single coordinate, depends on build-deep, runs in
  parallel with test-deep.  Generates Dokka API docs, builds the mkdocs
  site, and checks for broken links with djlint.

- test-wide: the remaining OS × Java coordinates.  Each wide coordinate
  runs `./gradlew test` and depends only on validate-inputs, so all wide
  jobs start immediately — they don't wait for build-deep.

- detekt / ktlint: depend only on validate-inputs (source checkout is
  sufficient), so they also run in parallel with build-deep.

- coverage-verification is eliminated as a separate job; coverage report
  generation and threshold checking are folded into test-deep via the new
  `testWithCoverage` Gradle task (test + testCodeCoverageReport +
  testCodeCoverageVerification).

validate_inputs.py computes and exposes the deep coordinate (coverage_os,
coverage_java) and the wide matrix (wide_matrix JSON array, has_wide flag)
as GITHUB_OUTPUT values, so the workflow never hardcodes OS or Java version
assumptions.

`./gradlew clean` has been removed from build-deep.  On a fresh CI runner
the build directory is already empty, so clean is a no-op.  More
importantly, Gradle's build cache is content-addressed: cache keys are
derived from all task inputs (source files, classpaths, JVM args), so a
cached result is only reused when the inputs are identical to a prior
passing run.  We trust that mechanism to ensure correctness rather than
forcing unconditional recompilation.


rstata:
Just pushed a fix: `GRADLE_OPTS_EXTRA` was silently ignored by gradlew; renamed to `GRADLE_OPTS` throughout.

PR #308, opened 2026-03-28
build(fix): Restore detekt/ktlint CI and fix build-scan upload steps

rstata:
## Summary

- Adds `detekt` and `ktlintCheck` as workspace-level orchestration tasks in `orchestration.gradle.kts`, following the existing pattern for `build`, `check`, `test`, etc. After the March 2026 build restructuring moved all source modules into the `core` and `gradle-plugins` included builds, these tasks were no longer visible at the root project level, causing CI to fail with \"Task not found\" on every run.
- Fixes the \"Upload build scan URLs\" step in all five jobs (`build`, `test`, `detekt`, `ktlint`, `coverage-verification`) by adding `always() &&` to the `if` condition. Without it, GitHub Actions' implicit `success() &&` prefix caused the upload to be skipped whenever a prior step had failed — exactly the situation where the scan URL is most needed.

## Test plan

- [ ] Verify `./gradlew detekt` and `./gradlew ktlintCheck` resolve and run across `core` and `gradle-plugins` modules
- [ ] Confirm CI detekt and ktlint jobs pass on this branch
- [ ] Build-scan upload fix was validated on `rstata/viaduct` branch `test-buildscan`: both "Upload build scan URLs" steps showed `success` (previously `skipped`) in a failing detekt/ktlint run

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -159,7 +159,7 @@ jobs:
           path: build/
 
       - name: Upload build scan URLs
-        if: ${{ steps.save_build_scan_url.conclusion == 'success' }}
+        if: ${{ always() && steps.save_build_scan_url.conclusion == 'success' }}
         uses: actions/upload-artifact@v6
         with:
           name: build-scan-urls-build-${{ matrix.java }}-${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


geovannefduarte (approved):
LGTM

PR #315, opened 2026-04-10
refactor(ci): extract alert formatting into composite action and isolate Python tests

geovannefduarte:
## Description

Three workflows duplicated the same inline shell pattern for formatting CI failure alerts (build JSON with jq, pipe to `format_alert.py`, write to `$GITHUB_OUTPUT`).
This PR extracts that pattern into a reusable composite action and moves Python script tests into their own lightweight job so they run in seconds without waiting for the Gradle build pipeline.

**Key Changes**

- `.github/actions/collect-failure-info/action.yml` — new composite action that accepts a `jobs_json` array and context inputs, constructs the alert payload, and calls `format_alert.py`. Follows the same caller-must-checkout pattern as `setup-build`.
- `.github/workflows/ci-manual-trigger.yml` — split the monolithic `format-alerts` step into a job-filtering step and a composite action call, removing ~20 lines of inline shell.
- `.github/workflows/periodic-green-check.yml` — replaced inline staleness alert formatting with the composite action.
- `.github/workflows/post-alerts.yml` — replaced inline test alert formatting with the composite action.
- `.github/workflows/build-and-test.yml` — added a standalone `python-tests` job (no JVM/Gradle, runs in parallel) and removed Python `test` steps from the Gradle test matrix job, eliminating 18 redundant step executions per CI run (3 steps x 6 matrix cells).

CI and gradle build

- `actionlint` passes on all modified and new workflow files
- Python unit tests pass locally via `python3 -m unittest discover -s .github/scripts/tests`
- Alert output format is unchanged (composite action produces identical JSON payloads to the inline shell it replaces)

### DIFF
--- a/.github/actions/collect-failure-info/action.yml
+++ b/.github/actions/collect-failure-info/action.yml
@@ -1,0 +1,64 @@
+name: Collect Failure Info
+description: Formats CI failure metadata into an alert message using format_alert.py
+
+inputs:
+  jobs_json:
+    description: 'JSON array of failed jobs, each with "name" and "run_id" fields'
+    required: true
+  branch:
+    description: 'Branch name'
+    required: true
+  server_url:
+    description: 'GitHub server URL'
+    required: true
+  repository:
+    description: 'Full repository name (owner/repo)'
+    required: true
+  sha:
+    description: 'Commit SHA (optional)'
+    required: false
+    default: ''
+  actor:
+    description: 'GitHub actor (optional)'
+    required: false
+    default: ''
+
+outputs:
+  text:
+    description: 'Formatted alert text'
+    value: ${{ steps.fmt.outputs.text }}
+
+runs:
+  using: composite
+  steps:
+    - name: Format alert
+      id: fmt
+      shell: bash
+      env:
+        JOBS_JSON: ${{ inputs.jobs_json }}
+        BRANCH: ${{ inputs.branch }}
+        SERVER_URL: ${{ inputs.server_url }}
+        REPOSITORY: ${{ inputs.repository }}
+        SHA: ${{ inputs.sha }}
+        ACTOR: ${{ inputs.actor }}
+      run: |
+        # Build the full payload — format_alert.py handles empty sha/actor
+        # gracefully via .get() + truthiness checks.
+        text=$(jq -n \
+          --arg branch "$BRANCH" \
+          --arg server_url "$SERVER_URL" \
+          --arg repository "$REPOSITORY" \
+          --arg sha "$SHA" \
+          --arg actor "$ACTOR" \
+          --argjson jobs "$JOBS_JSON" \
+          '{ branch: $branch, server_url: $server_url, repository: $repository,
+             jobs: $jobs }
+           + (if $sha  != "" then { sha: $sha }   else {} end)
+           + (if $actor != "" then { actor: $actor } else {} end)' | \
+          python3 .github/scripts/format_alert.py)
+
+        {
+          echo "text<<ALERT_EOF"
+          echo "$text"
+          echo "ALERT_EOF"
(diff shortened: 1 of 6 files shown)


rstata (changes requested):
## Review

Good refactoring overall — the composite action is well-structured, output references are preserved correctly across all callers, and isolating the Python tests into their own job (15s standalone vs duplicated across 6 matrix cells) is a nice efficiency win.

One bug and one nit.

### Bug: multi-line `jobs_json` will break `$GITHUB_OUTPUT` in `ci-manual-trigger.yml`

Lines 72/75/78 call `jq` without `-c`, so the output is pretty-printed (multi-line). Line 80 then writes it as:

```bash
echo "jobs_json=$jobs_json" >> "$GITHUB_OUTPUT"
```

The `echo "key=value"` format only works for single-line values. When a CI job actually fails, the multi-line JSON will be truncated at the first newline, and the composite action will receive invalid JSON.

CI is green on this PR because the `format-alerts` job only runs on failure — this path was not exercised.

**Fix**: add `-c` (compact) to the jq calls:

```bash
jobs_json=$(echo "$jobs_json" | jq -c --arg name "Build and Test" --arg run_id "$BAT_RUN_ID" '. + [{"name": $name, "run_id": $run_id}]')
```

(Same for the SDT and BCV lines.)

### Nit: display name change

"Standalone Tests" was renamed to "Demo App Tests" on line 75. Intentional? Fine either way, just flagging since it'll change the alert text.

rstata:
from claude:

>   Bug found: `ci-manual-trigger.yml` line 80 writes multi-line jq output (pretty-printed JSON) to `$GITHUB_OUTPUT` using the single-line echo "key=value" format.  This will silently truncate the jobs_json value when a CI job actually fails, feeding invalid JSON to the composite action.  The fix is adding -c to the three jq calls so they produce compact single-line JSON.  This wasn't caught by CI because the format-alerts job only runs on failure.

If you confirm this is a bug is there a regression test you can add for it?

geovannefduarte:
> ## Review
> Good refactoring overall — the composite action is well-structured, output references are preserved correctly across all callers, and isolating the Python tests into their own job (15s standalone vs duplicated across 6 matrix cells) is a nice efficiency win.
> 
> One bug and one nit.
> 
> ### Bug: multi-line `jobs_json` will break `$GITHUB_OUTPUT` in `ci-manual-trigger.yml`
> Lines 72/75/78 call `jq` without `-c`, so the output is pretty-printed (multi-line). Line 80 then writes it as:
> 
> ```shell
> echo "jobs_json=$jobs_json" >> "$GITHUB_OUTPUT"
> ```
> 
> The `echo "key=value"` format only works for single-line values. When a CI job actually fails, the multi-line JSON will be truncated at the first newline, and the composite action will receive invalid JSON.
> 
> CI is green on this PR because the `format-alerts` job only runs on failure — this path was not exercised.
> 
> **Fix**: add `-c` (compact) to the jq calls:
> 
> ```shell
> jobs_json=$(echo "$jobs_json" | jq -c --arg name "Build and Test" --arg run_id "$BAT_RUN_ID" '. + [{"name": $name, "run_id": $run_id}]')
> ```
> 
> (Same for the SDT and BCV lines.)
> 
> ### Nit: display name change
> "Standalone Tests" was renamed to "Demo App Tests" on line 75. Intentional? Fine either way, just flagging since it'll change the alert text.

"Demo App Tests" was already the name on main before this PR — no change was made here. The original format-alerts job in ci-manual-trigger.yml used this name since it was introduced.

For the multiline, great catch. I added `-c` flag to all three jq calls so `jobs_json` is always single-line. Added regression tests in `TestJobsJsonCompactness` to prevent this from regressing.

rstata (approved):

PR #316, opened 2026-04-10
feat(ci): add publish-branch.yml atomic workflow for snapshot and release publication

geovannefduarte:
## Description

Adds the core atomic publication workflow that publishes Viaduct artifacts from a given branch ref in either snapshot or release mode. This is the foundation that downstream orchestration workflows (nightly-build, demoapps-nightly-check, and eventually the release flow) will call to close the mavenLocal/Maven Central gap — where CI can be green but real publication fails.

**Key Changes**

- **New workflow: `.github/workflows/publish-branch.yml`**
  - Dual trigger: `workflow_dispatch` (manual) + `workflow_call` (reusable)
  - Inputs: `mode` (snapshot/release), `ref`, `run_checks`, `confirm_versions`
  - 4-job pipeline: `validate` → `confirm-versions` (optional) / `check` (optional) → `publish`
  - Validation gate via `validate_release_state.py` (Slice 8) — fails early on invalid branch/version state
  - Version confirmation via `confirmDemoAppVersions` Gradle task (Slice 6)
  - Snapshot mode publishes to Sonatype OSSRH only; release mode publishes to both Maven Central and Gradle Plugin Portal
  - `RELEASE` env var controls GPG signing (same mechanism as `release.yml`)
  - Outputs `source_sha`, `version`, `snapshot_tag` for orchestration consumers
  - Least-privilege permissions (`contents: read`), concurrency group with `cancel-in-progress: false`
- **No modifications to existing files**

- `actionlint` passes locally with zero errors
- Triggered `publish-branch.yml` in snapshot mode on `main` via `gh workflow run` — validate, confirm-versions, check, and publish jobs all succeeded; snapshot artifacts visible on Sonatype OSSRH

### DIFF
--- a/.github/workflows/publish-branch.yml
+++ b/.github/workflows/publish-branch.yml
@@ -134,6 +134,7 @@ jobs:
       && (needs.confirm-versions.result == 'success' || needs.confirm-versions.result == 'skipped')
       && (needs.check.result == 'success' || needs.check.result == 'skipped')
     runs-on: ubuntu-latest
+    environment: ${{ inputs.mode == 'release' && 'release' || '' }}
     outputs:
       source_sha: ${{ steps.sha.outputs.source_sha }}
     steps:
@@ -152,8 +153,6 @@ jobs:
 
       - name: Publish artifacts
         env:
-          VIADUCT_SONATYPE_USERNAME: ${{ secrets.VIADUCT_SONATYPE_USERNAME }}
-          VIADUCT_SONATYPE_PASSWORD: ${{ secrets.VIADUCT_SONATYPE_PASSWORD }}
           ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.VIADUCT_SONATYPE_USERNAME }}
           ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.VIADUCT_SONATYPE_PASSWORD }}
           ORG_GRADLE_PROJECT_signingKey: ${{ secrets.GPG_PRIVATE_KEY }}
--- a/.github/workflows/publish-branch.yml
+++ b/.github/workflows/publish-branch.yml
@@ -1,0 +1,180 @@
+name: Publish Branch
+
+on:
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Publication mode: snapshot or release"
+        required: true
+        type: choice
+        options:
+          - snapshot
+          - release
+      ref:
+        description: "Git ref to checkout and publish"
+        required: true
+        type: string
+      run_checks:
+        description: "Run ./gradlew check before publishing"
+        required: false
+        default: true
+        type: boolean
+      confirm_versions:
+        description: "Run ./gradlew confirmDemoAppVersions"
+        required: false
+        default: true
+        type: boolean
+
+  workflow_call:
+    inputs:
+      mode:
+        required: true
+        type: string
+      ref:
+        required: true
+        type: string
+      run_checks:
+        required: false
+        default: true
+        type: boolean
+      confirm_versions:
+        required: false
+        default: true
+        type: boolean
+    outputs:
+      source_sha:
+        description: "SHA of the published commit"
+        value: ${{ jobs.publish.outputs.source_sha }}
+      version:
+        description: "Version from VERSION file"
+        value: ${{ jobs.validate.outputs.effective_version }}
+      snapshot_tag:
+        description: "Suggested snapshot tag"
+        value: ${{ jobs.validate.outputs.snapshot_tag }}
+    secrets:
+      VIADUCT_GRADLE_PUBLISH_KEY:
+        required: false
+      VIADUCT_GRADLE_PUBLISH_SECRET:
+        required: false
+      VIADUCT_SONATYPE_USERNAME:
+        required: true
+      VIADUCT_SONATYPE_PASSWORD:
+        required: true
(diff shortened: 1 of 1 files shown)


rstata (dismissed):

rstata (commented):
## Review

Clean workflow. The structure is solid — validation gate, optional check/confirm-versions jobs, the `always() && ...` conditional on `publish`, concurrency group, minimal permissions.

A few observations:

### 1. No `environment` protection on publish

The existing `release.yml` uses `environment: release` (line 37), which gates publishes behind GitHub Environment protection rules (reviewers, wait timers, etc.). The new workflow has no `environment:` on the `publish` job.

If the intent is that orchestration callers handle gating, this is fine. But if anyone triggers this via `workflow_dispatch` in release mode, there's no approval gate before publishing to Maven Central and the Gradle Plugin Portal.

Could add conditionally:
```yaml
environment: ${{ inputs.mode == 'release' && 'release' || '' }}
```

### 2. Redundant env vars (nit)

`VIADUCT_SONATYPE_USERNAME` and `VIADUCT_SONATYPE_PASSWORD` are set in the `env:` block but nothing reads them — Gradle reads `ORG_GRADLE_PROJECT_mavenCentralUsername` / `ORG_GRADLE_PROJECT_mavenCentralPassword` (via vanniktech convention). The mapping is already there, so the `VIADUCT_SONATYPE_*` lines are dead weight.

### 3. `RELEASE` env var — this is actually better than the status quo

The existing `release.yml` → `publish_release.py` never sets `RELEASE=true`, so `viaduct-publishing.gradle.kts:43` defaults to `false` and GPG signing is skipped on releases. This workflow correctly sets `export RELEASE=true` in release mode. Nice fix.

### Looks good overall

- `validate_release_state.py` output names (`effective_version`, `snapshot_tag`, `is_release_branch`) match the job output declarations ✓
- `workflow_call` / `workflow_dispatch` input types are correct (`choice` for dispatch, `string` for call) ✓
- Concurrency group with `cancel-in-progress: false` is right for publish safety ✓
- `contents: read` permissions ✓
- Gradle Plugin Portal secrets correctly marked `required: false` (only needed for release mode) ✓

rstata:
more on the environment protection, i thinkwe should fix this:

> ● GitHub repos can define Environments (Settings → Environments). When a workflow job declares environment: release, GitHub enforces whatever protection rules are configured on that environment before the job runs. Common rules:
>
>  - Required reviewers — one or more people must click "Approve" in the Actions UI before the job starts
>  - Wait timer — e.g., a 5-minute delay to allow someone to cancel
>  - Branch restrictions — only certain branches can deploy to that environment
>  - Environment-scoped secrets — secrets that are only available to jobs targeting that environment
>
>  "Gating" just means "a checkpoint that blocks execution until some condition is met." In this case, if the release environment requires a reviewer, then a workflow_dispatch trigger in release mode would pause at the publish job and wait for someone to approve before publishing artifacts to Maven Central / Gradle Plugin Portal.
>
> The existing release.yml has environment: release on its publish job, so whatever protection rules are configured there apply. The new publish-branch.yml doesn't declare an environment, so those rules are bypassed — anyone with workflow dispatch permission can publish directly.
>
> Whether that matters depends on whether the orchestration callers are expected to handle that responsibility instead.

geovannefduarte:
> ## Review
> Clean workflow. The structure is solid — validation gate, optional check/confirm-versions jobs, the `always() && ...` conditional on `publish`, concurrency group, minimal permissions.
> 
> A few observations:
> 
> ### 1. No `environment` protection on publish
> The existing `release.yml` uses `environment: release` (line 37), which gates publishes behind GitHub Environment protection rules (reviewers, wait timers, etc.). The new workflow has no `environment:` on the `publish` job.
> 
> If the intent is that orchestration callers handle gating, this is fine. But if anyone triggers this via `workflow_dispatch` in release mode, there's no approval gate before publishing to Maven Central and the Gradle Plugin Portal.
> 
> Could add conditionally:
> 
> ```yaml
> environment: ${{ inputs.mode == 'release' && 'release' || '' }}
> ```
> 
> ### 2. Redundant env vars (nit)
> `VIADUCT_SONATYPE_USERNAME` and `VIADUCT_SONATYPE_PASSWORD` are set in the `env:` block but nothing reads them — Gradle reads `ORG_GRADLE_PROJECT_mavenCentralUsername` / `ORG_GRADLE_PROJECT_mavenCentralPassword` (via vanniktech convention). The mapping is already there, so the `VIADUCT_SONATYPE_*` lines are dead weight.
> 
> ### 3. `RELEASE` env var — this is actually better than the status quo
> The existing `release.yml` → `publish_release.py` never sets `RELEASE=true`, so `viaduct-publishing.gradle.kts:43` defaults to `false` and GPG signing is skipped on releases. This workflow correctly sets `export RELEASE=true` in release mode. Nice fix.
> 
> ### Looks good overall
> * `validate_release_state.py` output names (`effective_version`, `snapshot_tag`, `is_release_branch`) match the job output declarations ✓
> * `workflow_call` / `workflow_dispatch` input types are correct (`choice` for dispatch, `string` for call) ✓
> * Concurrency group with `cancel-in-progress: false` is right for publish safety ✓
> * `contents: read` permissions ✓
> * Gradle Plugin Portal secrets correctly marked `required: false` (only needed for release mode) ✓

Thanks for the review!
Both suggestions make sense — added the conditional `environment: release` gate on the publish job and removed the dead `VIADUCT_SONATYPE_*` env vars. Updated in the latest push.

PR #319, opened 2026-04-15
feat(ci): add ci-trigger, demoapps-nightly-check and nightly-build orchestration workflows

geovannefduarte:
## Description

Adds three orchestration workflows that compose the atomic workflows (push-demoapps, check-published-demoapps, publish-branch, demoapps-ci-check) into end-to-end automation flows. The centerpiece is the nightly check, which publishes snapshots to Sonatype, pushes demoapp code to the standalone viaduct-dev repos, waits for artifact visibility, and then verifies the standalone repos actually build — closing the gap where CI could be green via mavenLocal while real Maven Central publication was broken.

**Key Changes**

- `.github/workflows/ci-trigger.yml` — Fills the existing stub with a CI orchestrator that calls build-and-test, demoapps-ci-check, and bcv-api-check, with Slack/Discord alert formatting on failure. Uses workflow_dispatch/workflow_call only (push/PR triggers deferred to avoid duplicate runs with ci-manual-trigger.yml).
- `.github/workflows/demoapps-nightly-check.yml` — New end-to-end snapshot validation orchestrator: optional CI precheck → publish snapshots → push demoapps + wait for Sonatype visibility (in parallel) → verify standalone repos → cleanup tmp branches for ephemeral refs.
- `.github/workflows/nightly-build.yml` — Thin weekday cron wrapper (6am UTC) that delegates to demoapps-nightly-check with ref=main and run_ci_check=true. Contains no domain logic.
- `.github/impldocs/actions-arch.md` — Documents the three new orchestrators in the workflow taxonomy tables and adds flow diagrams for the nightly build and CI check paths.

**Static validation:**
- `actionlint` passes on all three workflow files (ci-trigger.yml, demoapps-nightly-check.yml, nightly-build.yml). Temporary stubs were created for slice 9/10 dependencies (publish-branch.yml, demoapps-ci-check.yml) to unblock validation since those PRs haven't merged yet.

**Local verification of alert pipeline:**
- The `format-alerts` jq construction in ci-trigger.yml was tested locally with simulated single-job and multi-job failure payloads piped through `format_alert.py`.
- Confirmed `jq -c` produces compact single-line JSON and the heredoc pattern writes multi-line alert text safely to `$GITHUB_OUTPUT`.

**Requires real GitHub Actions runs after dependencies merge (slices 9, 10, 11):**
- `gh workflow run ci-trigger.yml` — compare signal with ci-manual-trigger.yml
- `gh workflow run demoapps-nightly-check.yml -f ref=main -f run_ci_check=false` — full snapshot → push → verify loop
- `gh workflow run nightly-build.yml` — end-to-end cron path

  **Documentation:**
- `.github/impldocs/actions-arch.md` updated with orchestrator taxonomy entries, nightly build flow diagram, and CI check flow diagram

### DIFF
--- a/.github/impldocs/actions-arch.md
+++ b/.github/impldocs/actions-arch.md
@@ -63,7 +63,10 @@ An orchestrator composes atomics (and possibly inline jobs) into a higher-level 
 
 | Workflow | Purpose | Runs on |
 |---|---|---|
-| `ci-manual-trigger.yml` | Full CI suite: build-and-test + demoapp tests + API compat check | PR, merge to main, daily schedule (via periodic-green-check), manual |
+| `ci-manual-trigger.yml` | Full CI suite: build-and-test + standalone-demoapp-tests + API compat check | PR, merge to main, daily schedule (via periodic-green-check), manual |
+| `ci-trigger.yml` | Full CI suite: build-and-test + demoapps-ci-check + API compat check | manual, workflow_call (push/PR triggers to be added when `ci-manual-trigger.yml` is deleted) |
+| `demoapps-nightly-check.yml` | End-to-end snapshot validation: publish → push → wait → verify → cleanup | manual, workflow_call |
+| `nightly-build.yml` | Nightly cron wrapper, delegates to `demoapps-nightly-check.yml` | weekday schedule (6am UTC), manual |
 | `periodic-green-check.yml` | Scheduled CI health check + branch staleness detection | daily schedule, manual |
 
 ### Orchestrator Helpers
@@ -261,6 +264,53 @@ periodic-green-check.yml  [orchestrator]
                format-alert --> send-staleness-alert --> post-alerts.yml [helper] --> Slack + Discord
 ```
 
+### Nightly Build (6am UTC weekdays)
+
+```
+schedule / manual dispatch
+  |
+  v
+nightly-build.yml  [orchestrator, thin wrapper]
+  |
+  v
+demoapps-nightly-check.yml  [orchestrator]
+  |
+  |--- ci-precheck (if run_ci_check)
+  |      v
+  |    demoapps-ci-check.yml  [atomic]
+  |
+  |--- publish-snapshot
+  |      v
+  |    publish-branch.yml  [atomic, mode=snapshot]
+  |
+  |--- push-demoapps (parallel)       wait-for-publication (parallel)
+  |      v                               v
+  |    push-demoapps.yml  [atomic]     (poll Sonatype until 200)
+  |
+  |--- check-demoapps
+  |      v
+  |    check-published-demoapps.yml  [atomic]
+  |
+  '--- cleanup (if tmp/* branch)
+         delete tmp branches from viaduct-dev/* repos
+```
+
+### CI Check (via `ci-trigger.yml`)
+
+```
+manual dispatch / workflow_call
+  |
+  v
+ci-trigger.yml  [orchestrator]
+  |
+  |--- build-and-test.yml  [atomic]
+  |--- demoapps-ci-check.yml  [atomic]
+  |--- bcv-api-check.yml  [atomic]
+  |
+  '--- [if any atomic failed && send_alerts]
+         format-alerts --> send-alerts --> post-alerts.yml [helper] --> Slack + Discord
+```
+
 ### Manual Dispatch
 
 Every workflow supports `workflow_dispatch` so it can be run by hand independently. This serves two purposes: **testing** (validate a workflow change on a branch before merging) and **on-demand execution** (run a check or suite without waiting for its automatic trigger).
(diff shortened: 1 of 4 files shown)


.github/impldocs/actions-arch.md
```diff
@@ -63,7 +63,10 @@ An orchestrator composes atomics (and possibly inline jobs) into a higher-level
 
 | Workflow | Purpose | Runs on |
 |---|---|---|
-| `ci-manual-trigger.yml` | Full CI suite: build-and-test + demoapp tests + API compat check | PR, merge to main, daily schedule (via periodic-green-check), manual |
+| `ci-manual-trigger.yml` | Full CI suite: build-and-test + standalone-demoapp-tests + API compat check | PR, merge to main, daily schedule (via periodic-green-check), manual |
```
rstata, line 66: a couple of lines above it says **Current orchestratons** -- i think that's a typo, it should be atomics it think.

geovannefduarte: This is correct, `ci-manual-trigger.yml` is an orchestrator
It composes `build-and-test`, `demoapp-tests` and `bcv-api-check` into one suite and owns the alert logic

> ### Orchestrator Workflows
> 
> An orchestrator composes atomics (and possibly inline jobs) into a higher-level automation flow. It should contain little or no domain logic that really belongs in an atomic.

.github/workflows/demoapps-nightly-check.yml
```diff
@@ -0,0 +1,143 @@
...
+          DEST_BRANCH: ${{ needs.push-demoapps.outputs.destination_branch }}
+          DEMOAPP: ${{ matrix.demoapp }}
+        run: |
+          set -euo pipefail
+          echo "Deleting branch '${DEST_BRANCH}' from viaduct-dev/${DEMOAPP}..."
+          STATUS=$(curl -s -o /dev/null -w "%{http_code}" \
+            -X DELETE \
+            -H "Authorization: token ${GH_TOKEN}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/viaduct-dev/${DEMOAPP}/git/refs/heads/${DEST_BRANCH}")
```
rstata: can't we use the `gh` tool in our actions?

.github/workflows/demoapps-nightly-check.yml
```diff
@@ -0,0 +1,143 @@
...
+    name: Cleanup ${{ matrix.demoapp }}
+    steps:
+      - name: Delete temporary branch
+        env:
+          GH_TOKEN: ${{ secrets.VIADUCT_GRAPHQL_GITHUB_ACCESS_TOKEN }}
+          DEST_BRANCH: ${{ needs.push-demoapps.outputs.destination_branch }}
+          DEMOAPP: ${{ matrix.demoapp }}
+        run: |
+          set -euo pipefail
+          echo "Deleting branch '${DEST_BRANCH}' from viaduct-dev/${DEMOAPP}..."
```
rstata: deletions are dangerous, so two things would help reduce risk:

- Output the current SHA of the branch as well as the branch name so we could recover if we need to
- Double-check here that the DEST_BRANCH starts with `tmp/*`

.github/workflows/demoapps-nightly-check.yml
```diff
@@ -0,0 +1,143 @@
...
+    uses: ./.github/workflows/check-published-demoapps.yml
+    with:
+      ref: ${{ needs.push-demoapps.outputs.destination_branch }}
+
+  # ---------------------------------------------------------------------------
+  # Delete temporary branches in viaduct-dev/* repos (ephemeral refs only)
+  # ---------------------------------------------------------------------------
+  cleanup:
+    needs: [push-demoapps, check-demoapps]
+    if: always() && needs.push-demoapps.result == 'success' && startsWith(needs.push-demoapps.outputs.destination_branch, 'tmp/')
```
rstata: since check-demoapps needs push-demoapps, i don't think we need both of these `needs`

also, i think you should run this even if the push-demoapps failed, because the branch could have been created but the copy could've failed.  in the code below you can confirm that the branch exists and exit successfully if it doesn't -- that will capture the case that `push-demoapps` fails. before it even creates the branch.

.github/workflows/push-demoapps.yml
```diff
@@ -0,0 +1,97 @@
...
+            DEST_BRANCH="tmp/${REF}"
+          fi
+
+          if [ -z "$DEST_BRANCH" ]; then
+            echo "::error::Failed to compute destination branch"
+            exit 1
+          fi
+
+          echo "destination_branch=${DEST_BRANCH}" >> "$GITHUB_OUTPUT"
+          echo "Destination branch: ${DEST_BRANCH}"
```
rstata, line 69: i feel like i've seen something similar in other PRs.  is there a reusable (and testable) python script we can use for destination-branch computation?  (if i'm hallucinating and this isn't being repeated then i think inline shell is okay here).

geovannefduarte: Yeah, this mapping comes from the RFC/plan docs, so that’s likely where the familiarity is coming from

That said, I checked the codebase and this is the only place it’s implemented, no shared Python script or reuse elsewhere.
Given it’s a single use and pretty straightforward, inline shell feels reasonable. If we end up reusing it, we can extract it into a shared script.

.github/workflows/demoapps-nightly-check.yml
geovannefduarte: Good call, switched from curl to `gh api`. It's pre-installed on Actions runners and handles headers/auth
automatically. We set `GH_TOKEN` to the cross-repo PAT since the default `GITHUB_TOKEN` only covers airbnb/viaduct

.github/workflows/demoapps-nightly-check.yml
geovannefduarte: Agreed on both. The cleanup step now:

1. Hard-fails if `DEST_BRANCH` doesn't start with `tmp/` (defense-in-depth against future `if:` edits)
2. Fetches and logs the branch SHA before deleting, with a recovery command in the output: `git update-ref refs/heads/<branch> <sha>`

.github/workflows/demoapps-nightly-check.yml
geovannefduarte: Good point about partial failures — Copybara could create branches in some repos then fail on others.

Removed the `push-demoapps.result == 'success'` requirement so cleanup runs regardless. The script now checks if the branch exists first and exits cleanly if it doesn't, which handles the case where push-demoapps fails before creating any branches.

On the `needs` list: I kept both `push-demoapps` and `check-demoapps`. We need `push-demoapps` in `needs` to access its `destination_branch` output (GitHub Actions only exposes outputs from jobs in your `needs` list), and `check-demoapps` ensures cleanup runs last after verification completes.

.github/impldocs/actions-arch.md
rstata: ok

rstata (approved):

PR #321, opened 2026-04-17
fix: use Central Portal snapshot endpoint for snapshot publications

geovannefduarte:
## Description

Snapshot publications silently fail because the vanniktech `gradle-maven-publish-plugin` v0.34.0 uses Central Portal by default but doesn't route `-SNAPSHOT` versions to the snapshot endpoint. The `publishToMavenCentral` Gradle task succeeds (exit 0) but artifacts never reach any snapshot repository — the nightly build's `wait-for-publication` step times out after 15 minutes.

This was discovered during the first real run of the nightly snapshot validation pipeline (`nightly-build.yml`).

**Key Changes**

- `build-logic/src/main/kotlin/conventions/viaduct-publishing.gradle.kts` — Add a `snapshots` Maven repository when `RELEASE != true`, pointing to `central.sonatype.com/repository/maven-snapshots/`. Only registered for snapshot builds.
- `build-logic/shared/build.gradle.kts` — Same snapshot repository for `build-shared`, which has its own publishing block outside the convention plugin.
- `build-logic/src/main/kotlin/buildroot/orchestration.gradle.kts` — Add `publishToSnapshots` orchestration task that aggregates `publishAllPublicationsToSnapshotsRepository` across all subprojects and participating included builds, following the same pattern as `publishToMavenCentral`.
- `build.gradle.kts` (root) — Wire `build-shared` from `build-logic` into `publishToSnapshots`, matching the existing `publishToMavenCentral` wiring.
- `.github/workflows/publish-branch.yml` — Snapshot mode now calls `publishToSnapshots` instead of `publishToMavenCentral`. Release mode is unchanged.
- `.github/workflows/demoapps-nightly-check.yml` — Update polling URL to Central Portal snapshot endpoint.
- `demoapps/*/settings.gradle.kts` (5 files) — Update `USE_VIADUCT_SNAPSHOT_REPO` consumer URL from OSSRH to Central Portal.

Local validation: `RELEASE=false ./gradlew publishToSnapshots --dry-run` confirms the task graph resolves correctly with all modules included (api, bom, runtime, test-fixtures, build-shared, core/*, gradle-plugins/*).

### DIFF
--- a/build-logic/src/main/kotlin/conventions/viaduct-publishing.gradle.kts
+++ b/build-logic/src/main/kotlin/conventions/viaduct-publishing.gradle.kts
@@ -41,7 +41,10 @@ apply(from = generateSequence(gradle) { it.parent }.last()
 
 mavenPublishing {
     val isRelease = providers.environmentVariable("RELEASE").orElse("false").get().toBoolean()
-    publishToMavenCentral(automaticRelease = true)
+    publishToMavenCentral(
+        host = if (isRelease) SonatypeHost.CENTRAL_PORTAL else SonatypeHost.S01,
+        automaticRelease = true
+    )
     if (isRelease) {
         signAllPublications()
     }
--- a/build-logic/src/main/kotlin/conventions/viaduct-publishing.gradle.kts
+++ b/build-logic/src/main/kotlin/conventions/viaduct-publishing.gradle.kts
@@ -41,10 +41,7 @@ apply(from = generateSequence(gradle) { it.parent }.last()
 
 mavenPublishing {
     val isRelease = providers.environmentVariable("RELEASE").orElse("false").get().toBoolean()
-    publishToMavenCentral(
-        host = if (isRelease) SonatypeHost.CENTRAL_PORTAL else SonatypeHost.S01,
-        automaticRelease = true
-    )
+    publishToMavenCentral(automaticRelease = true)
     if (isRelease) {
         signAllPublications()
     }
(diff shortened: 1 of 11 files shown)


build-logic/src/main/kotlin/conventions/viaduct-publishing.gradle.kts
```diff
@@ -55,6 +55,29 @@ mavenPublishing {
...
+// can publish snapshots to OSSRH. The workflow calls this task instead of `publishToMavenCentral`
+// when in snapshot mode.
+run {
+    val isRelease = providers.environmentVariable("RELEASE").orElse("false").get().toBoolean()
+    if (!isRelease) {
+        publishing {
+            repositories {
+                maven {
+                    name = "sonatypeSnapshots"
+                    url = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")
```
gokhan-ozgozen: s01.oss.sonatype.org is being shut down June 30, 2025 — Sonatype has announced OSSRH end-of-life: https://central.sonatype.org/pages/ossrh-eol/

If I am not mistaken, Sonatype Central Portal now supports SNAPSHOT publishing natively [here](https://central.sonatype.org/publish/publish-portal-snapshots/). Perhaps it might be more future proof to check if a newer plugin version handles it correctly?

geovannefduarte: Great catch!

I checked `v0.35.0` and `v0.36.0` of the vanniktech plugin, which do have built-in snapshot routing, but both require Gradle 9+ / Kotlin 2.2+ which we're not on yet (Gradle 8.11, Kotlin 1.9.25). So I've updated the PR to use the Central Portal snapshot endpoint (`central.sonatype.com/repository/maven-snapshots/`) directly instead of OSSRH, same architecture but future-proof URL.
When we eventually bump Gradle/Kotlin, we can simplify by upgrading the plugin and dropping the manual snapshot repo plumbing.

gokhan-ozgozen (approved):

PR #323, opened 2026-04-19
chore: delete publish_release.py, replace with direct Gradle invocations

geovannefduarte:
## Description

Deletes `publish_release.py` and replaces the call in `release.yml` with direct Gradle invocations, matching the pattern used by `publish-branch.yml`.

**Audit summary:** Nothing useful is lost. The Plugin Portal version check (`check_version_published`) is unnecessary — the Gradle Plugin Portal rejects duplicate versions with a clear error. All other functionality (credential mapping, release/snapshot routing, Gradle task invocations) is replicated directly in the YAML `run:` step.

**Key Changes**

- `.github/scripts/publish_release.py` — deleted
- `.github/workflows/release.yml` — replaced `python3 publish_release.py` with direct Gradle commands using the same env var mapping pattern as `publish-branch.yml`

- `grep -r "publish_release" .github/` returns nothing
- `release.yml` remains functional as a deprecated fallback with identical publish behavior


.github/workflows/release.yml
```diff
@@ -68,11 +68,14 @@ jobs:
           chmod +x ./gradlew
           echo "✅ Gradle wrapper ready"
 
+      # Python is only needed by generate_changelog.py, which runs for releases, not snapshots.
```
rstata: Do we need to make this distinction?  Let's just install python and out dependencies for all cases just to simplify.

geovannefduarte: Agreed, removed the gates

rstata (approved):

PR #336, opened 2026-05-01
fix(ci): gate javaapi publications behind publishMinimal flag

geovannefduarte:
## Description

The `demoapps-ci-check` workflow fails because `publications/settings.gradle.kts` unconditionally includes `:javaapi-api` and `:javaapi-codegen` but `-PpublishMinimal` excludes their source modules from `core/settings.gradle.kts`. Without the source modules in the composite build, auto-substitution can't resolve `com.airbnb.viaduct.javaapi:api:INCLUDED` and the build fails.

**Key Changes**

- `publications/settings.gradle.kts` — Gate `:javaapi-api` and `:javaapi-codegen` includes behind the same `publishMinimal` property check used in `core/settings.gradle.kts` so both builds agree on which javaapi modules exist

- `actionlint` passes on the modified workflow

### DIFF
--- a/publications/settings.gradle.kts
+++ b/publications/settings.gradle.kts
@@ -18,5 +18,7 @@ include(":runtime")
 include(":buildtime")
 include(":test-fixtures")
 include(":bom")
-include(":javaapi-api")
-include(":javaapi-codegen")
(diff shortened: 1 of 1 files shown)


geovannefduarte:
Closed in favor of #338

PR #360, opened 2026-05-15
fix: Windows compatibility for codegen output and source-path matching

gokhan-ozgozen:
# fix: Windows compatibility for codegen output and source-path matching

## Description

Three Windows-only build/test issues, grouped by theme:

- **LF newline pins** — `STUtils`, `Schema2CsvCommand`, `ResolverParamsJsonCodec` were emitting CRLF via `System.lineSeparator()` on Windows, breaking byte-identical codegen output.
- **Path separator normalization** — `Extensions.tenant`, `ViaductSchemaExtensions.tenantModule`, and `ZipUtil` assumed `/` in regex inputs and ZIP entry paths. Normalize `\` → `/` (and use `invariantSeparatorsPath` for ZIP entries, per spec).
- **Test server startup timeout** — `ViaductServerTestHelper` waited 30s for cold-start, which is tight on Windows JVM warmup; bumped to 60s.

Symlink requirements are already fixed upstream via 562ed06b.

Published to localMaven and tested against standalone demoapps. `./gradlew check` passes on Windows.

### Demoapps Testing strategy

`viaduct` framework itself passes `./gradlew check`. Then published using `./gradlew publishToMavenLocal`.

Afterwards for each demoapp ( ie `starwars` ) 

1. clone the repo from `viaduct-dev` repo ( ie https://github.com/viaduct-dev/starwars )
2. run `set "USE_MAVEN_LOCAL=true" && gradlew.bat build -PviaductVersion=1.1.0-SNAPSHOT` and verify it succeeds. As I was working on this PR `1.1.0-SNAPSHOT` was the next targeted version, it might change but against local plugins demoapps work.


core/service/serve/src/test/kotlin/viaduct/serve/ViaductServerTestHelper.kt
```diff
@@ -22,8 +22,10 @@ class ViaductServerTestHelper : AutoCloseable {
             } catch (_: Exception) {
             }
         }.start()
-        port = waitFor(30_000) { if (server.actualPort != 0) server.actualPort else null }
-        waitFor(30_000) {
+        // 60s is generous for cold-start on slow CI / Windows where JVM warmup + class loading
+        // can take 20s+ before the port is even bound. Warm tests bind in <1s.
+        port = waitFor(60_000) { if (server.actualPort != 0) server.actualPort else null }
```
gokhan-ozgozen: The hardware I used for testing might not be sufficient and I observed some flakyness here so I bumped the wait time here. This one actually is not related to windows compatibility.

core/shared/viaductschema/src/testFixtures/kotlin/viaduct/graphql/schema/cli/Schema2CsvCommand.kt
```diff
@@ -250,18 +250,20 @@ data class CsvTable(
      * Creates two files: {baseName}s.csv and {baseName}dirs.csv
      */
     fun writeTo(outputDir: File) {
+        // Pin to LF so output is byte-identical across platforms (PrintStream.println
+        // would emit CRLF on Windows).
         val out = PrintStream(FileOutputStream(File(outputDir, "${baseName}s.csv")))
-        out.println(baseHeaders)
+        out.print(baseHeaders); out.print('\n')
```
fireboy1919: This kind of thing looks like a hack; environmental concerns are being built into our command structure.

Is there a more universal way to accomplish this?

gokhan-ozgozen: Yes, you are right. I have separated that concern and created a wrapper around `PrintStream` to address this. Please re-review.

core/shared/utils/src/testFixtures/kotlin/viaduct/utils/test/LfAssertions.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * assertEquals with line endings normalized to LF on both sides — makes string
+ * comparison more tolerant for cross-platform testing (e.g., CRLF on Windows).
+ */
+fun assertEqualsLf(
+    expected: String,
+    actual: String,
+    message: String? = null
+) = assertEquals(expected.lf(), actual.lf(), message)
```
fireboy1919: Rather than writing this yourself, how about using the [assertj function that does this](https://javadoc.io/doc/org.assertj/assertj-core/3.18.1/org/assertj/core/api/StringAssert.html)

geovannefduarte (approved):

Issue #366, opened 2026-05-30
Schema Scoping: Slice 1 — DSL surface + canonical SchemaScoping data class


xyu307:
This the first of 12 sub-issues breakdown of issue #361. It may refer to the original single consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

# PR 1 — DSL surface + canonical `SchemaScoping` data class

> **Status: updated 2026-05-28** — rewritten around the unified `SchemaScoping` data class. Module location updated after standup feedback. See [decisions.md](../decisions.md) D9–D11.

## Goal

Introduce the canonical `SchemaScoping` data class plus the DSL surface on `ViaductApplicationExtension` that accumulates state into it. **No validation, no plugin wiring beyond the extension property, no resource emission.** Just the data class, the DSL accumulation pattern, and ProjectBuilder tests for getters/setters.

## Why this is its own PR

The `SchemaScoping` data class is the foundation that PR 2, PR 3, PR 8, PR 9, and PR 10 all depend on. Landing it first — with no behavior beyond accumulation — keeps the architectural decision focused and lets reviewers reason about the API shape without confusion from validation logic or runtime semantics.

## Files touched

- `core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt` — the canonical data class. Sibling to `SchemaId.kt` in `service/api`.
- `gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt` — DSL surface
- `gradle-plugins/common/src/main/kotlin/viaduct/gradle/ScopedSchemaDefinition.kt` — internal wrapper
- `gradle-plugins/application/src/test/kotlin/viaduct/gradle/ViaductApplicationExtensionTest.kt` — ProjectBuilder tests
- Add a thin dep edge from `gradle-plugins/common` to `core/service/api` for the `SchemaScoping` import.

**Why `core/service/api`:** `SchemaScoping` is the build-to-runtime API contract — emitted by the Gradle plugin, consumed by `SchemaConfiguration` at runtime. `service/api` already holds the equivalent concrete public types (`SchemaId`, `ExecutionInput`, `GraphQLError`); this is the same shape. The runtime in `service/runtime` already depends on `service/api` and consumes for free. No new module needed.

### The canonical data class

```kotlin
package viaduct.service.api.scoping

import java.io.Serializable

data class SchemaScoping(
    val scopeUniverse: Set<String>,
[...]

PR #367, opened 2026-05-30
feat(scope): introduce SchemaScoping data class and DSL accumulation surface

xyu307:
Implements #366 — first slice of the #361 schema-scoping breakdown.

Foundational slice: introduces the canonical `SchemaScoping` data class and the DSL accumulation pattern on `ViaductApplicationExtension`. **No SDL validation, no resource emission, no runtime behavior change.** Follow-up slices layer those on. Duplicate-ID rejection on the DSL setters is in scope per [decisions.md D13](https://github.com/airbnb/viaduct/issues/361#issuecomment-4617603809) (the cross-slice decisions doc is attached to #361 as a comment).

### Why this is its own PR

`SchemaScoping` is the build-to-runtime API contract that #366's siblings (extension validation, SDL validation, manifest writer, runtime reader, builder wiring) all depend on. Landing the data class and the DSL shape first — with no behavioral payload beyond accumulate-with-dedup-rejection — keeps the architectural decision focused and lets reviewers reason about the API shape.

### What ships

- `SchemaScoping` data class in `core/service/api/.../scoping/`, sibling to `SchemaId.kt`. Fields, `isScoped` predicate, `EMPTY` constant, `CURRENT_VERSION`. `Serializable` for in-process Gradle plumbing; JSON wire format lands in a later slice.
- `ViaductApplicationExtension` gains `declaredSchemaScopes(Set<String>)` and `declaredScopedSchemas(vararg "id" to scopes)` DSL methods, backed by private storage. Both are **single-call immutable** — second invocation throws `GradleException`. Empty inputs and within-`declaredScopedSchemas` duplicate IDs are rejected with `GradleException`. Public surface is `Provider<SchemaScoping>` — re-evaluates on every `get()` to reflect current state.
- Internal `ScopedSchemaDefinition` wrapper around `Set<String>` (Gradle `MapProperty` value-type erasure workaround). Kept `internal`; projected back to `Set<String>` before reaching the public `SchemaScoping`.
- Tests cover the single-call DSL contract: 14 tests in `ViaductApplicationExtensionTest` plus 6 in `SchemaScopingTest`. All tests are behavioral — no source-file string inspection, no brittle line-count assertions.

### Out of scope (deferred to follow-up slices)

- ID-format / reserved-ID / subset / "absent → both absent" validation
- `@scope` directive injection toggle, SDL validation against the declared universe
- JSON manifest writer + runtime reader
- Builder wiring, demo-app migration, deprecation of raw-scope-set factories

### Reviewers should focus on

- **Module location.** `SchemaScoping` lives in `core/service/api/.../scoping/`, sibling to `SchemaId.kt`. Rationale: this is the build-to-runtime API contract; `service/api` already holds the equivalent concrete public types, and `service/runtime` consumes it transitively.
- **`abstract class @Inject` vs `open class(objects)`.** Kept the existing `open class` constructor to avoid touching the plugin's `extensions.create(..., objects)` call site. Easy to flip later if there's a preference; it's a local change.
- **`ScopedSchemaDefinition` kept `internal`.** It only exists to satisfy `MapProperty<K, V>`'s value-type erasure, and is projected away before reaching the public Provider type. Worth confirming this is the desired contract before any convention plugins start composing.
- **Duplicate-rejection error messages.** Both setters throw `GradleException` synchronously and the message includes the duplicate ID(s) so users can grep their build script. PR 2 will introduce the formalized error catalogue (codes + templated messages + remediation); these messages are placeholder-style for now and will be re-keyed then.

### Open questions, deviations, ambiguities of the spec

Three points worth a reviewer call. (A prior question on multi-call duplicate handling was resolved offline — see "Resolved" below.)

**1. `@get:Input` on the `schemaScoping` Provider omitted vs sample code.** The slice-doc sample shows:
```kotlin
@get:org.gradle.api.tasks.Input
val schemaScoping: Provider<SchemaScoping> = ...
```
This PR omits the annotation. Three reasons:
- Gradle only inspects `@Input` annotations on `Task` properties; on an extension, it is a no-op at runtime.
- None of the pre-existing extension properties on `ViaductApplicationExtension` (`grtPackageName`, `modulePackagePrefix`, `servePort`, `serveHost`) carry `@Input` either, so adding it only on the new property would be locally inconsistent.
- The slice doc's own "Out of scope" section says: *"Plumbing `schemaScoping` as `@Input` on `AssembleCentralSchemaTask` — PR 3."* The `@Input` lives on the task in slice 3, not on the extension.

Reading: the annotation in the sample is likely a small drafting inconsistency. Happy to add `@get:Input` back if a reviewer prefers cargo-cult-now-rename-later; otherwise it lands on the task in slice 3.

**2. Constructor shape: `open class(objects)` vs `abstract class @Inject`.** Also called out under "Reviewers should focus on" above. Sample is `abstract @Inject`; this PR keeps the existing `open class(objects)` constructor to minimize blast radius at the plugin call site (`extensions.create("viaductApplication", ViaductApplicationExtension::class.java, objects)` in `ViaductApplicationPlugin.kt:36`). Easy flip later.

**3. TestKit coverage of the DSL through a real `build.gradle.kts` — should slice 1 include it?** The slice spec's test plan only asks for `ProjectBuilder` tests, and this PR delivers exactly that — 14 tests instantiating `ViaductApplicationExtension(project.objects)` directly. The plugin application path (`extensions.create(...)` inside `ViaductApplicationPlugin.apply`) is **not** exercised by these tests. Specifically, three things remain unverified end-to-end:

- That `extensions.create("viaductApplication", ViaductApplicationExtension::class.java, objects)` still wires the new private `SetProperty` / `MapProperty` correctly (the constructor signature is unchanged, so this should be safe, but it's untested).
- That the duplicate-rejection `GradleException` actually propagates through Gradle's configuration phase as a clear build failure (vs. being re-wrapped in a way that hides the dup-id in the user-facing message).
- That a real `viaductApplication { declaredSchemaScopes(...); declaredScopedSchema(...) }` block in a consumer's `build.gradle.kts` behaves as expected.

There is already a TestKit pattern in the same module — `ViaductApplicationPluginFunctionalTest.kt` uses `GradleRunner` with a synthesized `settings.gradle.kts` + `build.gradle.kts` for exactly this kind of check, and [`gradle-plugins/CLAUDE.md`](../../blob/main/gradle-plugins/CLAUDE.md) explicitly lists "clear diagnostics for invalid or incomplete plugin configuration" and "confirming that correctly configured projects do not fail during configuration" as appropriate TestKit cases.

Three small cases would close the gap:
- happy-path DSL invocation succeeds at configuration time;
- duplicate scope ID via the real DSL fails the build with the dup-id in the output;
- duplicate scoped-schema ID via the real DSL fails the build with the dup-id in the output.

Two readings:
- **Add to slice 1**, because TestKit coverage of the DSL ergonomics is cheap (one file, ~3 cases), and it confirms the duplicate-rejection contract surfaces correctly in real builds — the most user-visible behavior in this slice.
- **Defer**, because the spec's test plan is explicit and the demoapp migration in slice 11 is the canonical end-to-end verification — adding TestKit now duplicates what slice 11 will catch anyway.

Leaning slightly toward "add to slice 1" since duplicate-rejection is the only error path users will hit from this slice, and a TestKit check makes the error message contract self-evident at review time. Open to either. Will add if reviewer agrees.

**Resolved — single-call immutable DSL (D13 rescinded).** Multi-call accumulation has been **rescinded** in response to follow-up review feedback. D13's "accumulate-with-dedup-rejection" is superseded by a single-call immutable contract: `declaredSchemaScopes(Set<String>)` and `declaredScopedSchemas(vararg "id" to scopes)` are each callable at most once, reject empty inputs, and reject duplicate scoped-schema IDs within the single varargs invocation. The singular `declaredScopedSchema(id, scopes)` is renamed to plural `declaredScopedSchemas` with varargs of `Pair`. Empty `Set<String>` as a per-entry value (the full-schema alias) remains allowed. Full spec change recorded at https://github.com/airbnb/viaduct/issues/366#issuecomment-4617579597; the cross-slice decisions doc attached to #361 still contains D13's original text for historical context.

## Test plan

- [x] `viaduct.service.api.scoping.SchemaScopingTest`: 6/6 passing (EMPTY, isScoped both branches, data-class equality, default version, Serializable round-trip)
- [x] `viaduct.gradle.ViaductApplicationExtensionTest`: 14/14 passing — covers the single-call DSL contract: default-empty, single-call universe + scoped-schemas, second-call rejection on both setters, empty-input rejection on both setters, within-call duplicate-ID rejection, empty-set value (full-schema alias), isScoped both branches, snapshot semantics — plus pre-existing-property regression and end-to-end DSL → SchemaScoping equivalence
- [x] Pre-existing `viaduct.gradle.ViaductApplicationPluginTest` ran in the same suite — no regressions
- [x] Local `./gradlew :core:service:api:test :gradle-plugins:common:test :gradle-plugins:application:test` → BUILD SUCCESSFUL
- [x] `./gradlew :core:service:api:apiCheck` → BUILD SUCCESSFUL (no BCV public-surface drift)
- [x] End-to-end `./gradlew :starwars:test` → BUILD SUCCESSFUL (plugin loads cleanly under the reshaped DSL)
- [ ] TestKit functional test exercising the DSL through a real `build.gradle.kts` — see Open Question #3

### DIFF
--- a/core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
+++ b/core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
@@ -1,7 +1,7 @@
 package viaduct.service.api.scoping
 
 import java.io.Serializable
-import viaduct.apiannotations.StableApi
+import viaduct.apiannotations.ExperimentalApi
 
 /**
  * Build-to-runtime contract describing the schema-scoping declarations of a Viaduct application.
@@ -18,7 +18,7 @@ import viaduct.apiannotations.StableApi
  *  set is an alias for the full schema.
  * @property version manifest schema version, used to evolve the on-disk JSON format.
  */
-@StableApi
+@ExperimentalApi
 data class SchemaScoping(
     val scopeUniverse: Set<String>,
     val scopedSchemas: Map<String, Set<String>>,
--- a/core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
+++ b/core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
@@ -1,0 +1,39 @@
+package viaduct.service.api.scoping
+
+import java.io.Serializable
+import viaduct.apiannotations.StableApi
+
+/**
+ * Build-to-runtime contract describing the schema-scoping declarations of a Viaduct application.
+ *
+ * Emitted by the `viaduct-application` Gradle plugin from the application's `declaredSchemaScopes` /
+ * `declaredScopedSchema` DSL, and consumed at runtime when a Viaduct instance resolves which scoped
+ * schemas to register. The canonical cross-process exchange format is the JSON manifest written
+ * into the application JAR; the [Serializable] implementation supports in-process Gradle plumbing
+ * (e.g. configuration-cache state) and is not the wire format for runtime consumption.
+ *
+ * @property scopeUniverse the complete set of scope IDs declared as valid for this application;
+ *  empty when the application does not opt into scoping.
+ * @property scopedSchemas mapping from declared scoped-schema ID to its scope set; an empty scope
+ *  set is an alias for the full schema.
+ * @property version manifest schema version, used to evolve the on-disk JSON format.
+ */
+@StableApi
+data class SchemaScoping(
+    val scopeUniverse: Set<String>,
+    val scopedSchemas: Map<String, Set<String>>,
+    val version: String = CURRENT_VERSION,
+) : Serializable {
+    /**
+     * Whether this application opts into scope-based filtering. The presence of a declared
+     * universe — not the contents of [scopedSchemas] — is the source of truth.
+     */
+    val isScoped: Boolean get() = scopeUniverse.isNotEmpty()
+
+    companion object {
+        const val CURRENT_VERSION: String = "1"
+
+        /** Convenience constant for the "no scoping declared" state. */
+        val EMPTY: SchemaScoping = SchemaScoping(emptySet(), emptyMap())
+    }
+}
--- a/core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
+++ b/core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
@@ -7,7 +7,7 @@ import viaduct.apiannotations.ExperimentalApi
  * Build-to-runtime contract describing the schema-scoping declarations of a Viaduct application.
  *
  * Emitted by the `viaduct-application` Gradle plugin from the application's `declaredSchemaScopes` /
- * `declaredScopedSchema` DSL, and consumed at runtime when a Viaduct instance resolves which scoped
+ * `declaredScopedSchemas` DSL, and consumed at runtime when a Viaduct instance resolves which scoped
  * schemas to register. The canonical cross-process exchange format is the JSON manifest written
  * into the application JAR; the [Serializable] implementation supports in-process Gradle plumbing
  * (e.g. configuration-cache state) and is not the wire format for runtime consumption.
--- a/gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt
+++ b/gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt
@@ -22,9 +22,12 @@ open class ViaductApplicationExtension(objects: ObjectFactory) {
     private val scopedSchemasProperty: MapProperty<String, ScopedSchemaDefinition> =
         objects.mapProperty(String::class.java, ScopedSchemaDefinition::class.java)
 
+    private var scopeUniverseDeclared = false
+    private var scopedSchemasDeclared = false
+
     /**
      * Canonical view of the application's schema-scoping declarations, derived from the
-     * `declaredSchemaScopes` and `declaredScopedSchema` DSL methods. Re-evaluated on each `get()`,
+     * `declaredSchemaScopes` and `declaredScopedSchemas` DSL methods. Re-evaluated on each `get()`,
      * so successive snapshots reflect the state at the moment of resolution.
      */
     val schemaScoping: Provider<SchemaScoping> =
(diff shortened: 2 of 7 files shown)


core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt
```diff
@@ -0,0 +1,39 @@
...
+ * into the application JAR; the [Serializable] implementation supports in-process Gradle plumbing
+ * (e.g. configuration-cache state) and is not the wire format for runtime consumption.
+ *
+ * @property scopeUniverse the complete set of scope IDs declared as valid for this application;
+ *  empty when the application does not opt into scoping.
+ * @property scopedSchemas mapping from declared scoped-schema ID to its scope set; an empty scope
+ *  set is an alias for the full schema.
+ * @property version manifest schema version, used to evolve the on-disk JSON format.
+ */
+@StableApi
```
geovannefduarte: Consider `@ExperimentalApi` here instead of `@StableApi` for now. `viaduct.service.api.*` is a BCV-tracked public package, so `@StableApi` commits us to "no change without a deprecation cycle" but this is slice 1 of the scoping series, and the companion surface are still in flux

xyu307: Good call — agreed and applied in commit eb1fc82e.

`@StableApi`'s KDoc explicitly frames it as the BCV-stability marker (and `build-logic/conventions/bcv-api.gradle.kts` confirms: `publicMarkers = [StableApi]`, `nonPublicMarkers += ExperimentalApi`). Since the companion surface picks up `parse` / `toJsonString` in PR 8/9, validation helpers in PR 2, and a possibly-reshaped `version` field along the way, anchoring slice 1 to "no change without a deprecation cycle" is premature.

Flipped `SchemaScoping` to `@ExperimentalApi`, added `@OptIn(ExperimentalApi::class)` at the `ViaductApplicationExtension` consumer and the two test classes, and regenerated `core/service/api/api/api.api` (which correctly removes the `SchemaScoping` entries — `@ExperimentalApi` is `nonPublicMarkers`, so BCV no longer tracks it). Plan to promote back to `@StableApi` once the manifest round-trip is end-to-end verified — most likely alongside the demoapp migration in slice 11.

Verification on the rebased branch:
- targeted unit tests: `SchemaScopingTest` 6/6, `ViaductApplicationExtensionTest` 12/12, pre-existing `ViaductApplicationPluginTest` 11/11
- `:core:service:api:apiCheck` green
- end-to-end `:starwars:test` BUILD SUCCESSFUL (169 tasks across the composite)

xyu307: Above is generated by Claude in response. It also rebased against the latest `main`. See if any other change is required.

geovannefduarte (dismissed):

gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt
```diff
@@ -1,9 +1,70 @@
...
+import viaduct.service.api.scoping.SchemaScoping
+
+open class ViaductApplicationExtension(objects: ObjectFactory) {
     /** Kotlin package name for generated GRT classes. */
     val grtPackageName = objects.property(String::class.java).convention("viaduct.api.grts")
 
     /** Kotlin package name prefix for all modules. */
     val modulePackagePrefix = objects.property(String::class.java)
+
+    private val scopeUniverseProperty: SetProperty<String> =
```
rstata, line 19: can we add some validation to this so that you can't explicitly set it to empty?  this way "never set" means "default to empty" means "no scoping," where "set" means "there is scoping" implies "there's at least one scope in the scope universe"

rstata:
Looking at this code, i can see the original spec requires mutability for the declared scoped-schemas where we don't need it.  So let's change the syntax to this:

```kotlin
declaredScopedSchemas(
    "FULL_ALIAS" to emptySet(),
    "PUBLIC_API" to setOf("public"),
    "INTERNAL_API" to setOf("public", "internal"),
    "LEGACY_INTERNAL_API" to setOf("public", "internal")
)
```

gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt
```diff
@@ -1,9 +1,72 @@
...
+            )
+        }
+
+    /**
+     * Adds [scopes] to the declared scope universe. May be called multiple times; calls accumulate
+     * via set union. A scope ID that already appears in the universe from an earlier call is
+     * rejected with a [GradleException] naming the duplicate, so that convention plugins and
+     * application code can't silently shadow each other.
+     */
+    fun declaredSchemaScopes(scopes: Set<String>) {
```
rstata, line 48: given the change recommended below that we have just one `declaredScopedSchemas` declaration let's also have just one `declareSchemaScopes` declaration.

xyu307:
@rstata @geovannefduarte I will treat above as clarification and change of specs and instruct agents to record it on  issue #366 and make changes.

gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt
xyu307: Ok. In the similar spirit, I am adding rejection when declaredScopedSchemas() is called with zero varargs as default meant "no scoping" and set means scoping with at least one schema declared. Hope this makes sense without flaws.

xyu307:
Thanks @rstata — all three pieces of feedback applied in `c0ec0a36` as a separate commit on top of the prior PR head:

1. **`declaredScopedSchemas(vararg "id" to scopes)`** — the singular `declaredScopedSchema(id, scopes)` is renamed to plural and reshaped to `vararg entries: Pair<String, Set<String>>`, matching the syntax in your comment. Single-call: a second invocation throws `GradleException`. Duplicate IDs within the single varargs invocation also throw.
2. **Non-empty `declaredSchemaScopes`** — empty `Set<String>` is now rejected at the setter with `GradleException`, directing the caller to omit the call entirely if no scoping is intended.
3. **Single-call `declaredSchemaScopes`** — second invocation throws `GradleException`, matching the single-call constraint on `declaredScopedSchemas`.

For symmetry with (2), an empty varargs list to `declaredScopedSchemas()` is also rejected.

The empty `Set<String>` *value* per entry (e.g. `"FULL_ALIAS" to emptySet()` as a full-schema alias) remains allowed — that's the documented value contract, not a structural emptiness.

`ViaductApplicationExtensionTest` now has 14 tests (was 12): one obsolete duplicate-across-multi-call test removed, three new ones added covering second-call rejection on `declaredScopedSchemas`, empty-set rejection on `declaredSchemaScopes`, and empty-entries-list rejection on `declaredScopedSchemas`. Targeted suite, BCV `apiCheck`, and the `starwars` end-to-end test all pass.

Spec change recorded on issue #366.

---

_Above is generated by Claude Code._

xyu307:
**Rebased onto current `origin/main`** (now includes `d9bf5b18` and 6 other upstream commits; latest engine work — QueryPlan refactor, FieldSelectivity SPI, remote-resolver enablement). Rebase was clean (no conflicts; no path overlap with this PR's files). New PR head: `2d47b5ab`.

**Verification post-rebase:**
- `viaduct.gradle.ViaductApplicationExtensionTest`: 14/14 ✓
- `viaduct.service.api.scoping.SchemaScopingTest`: 6/6 ✓
- `:core:service:api:apiCheck` (BCV) ✓
- `:gradle-plugins:common:test` and `:gradle-plugins:application:test` ✓

**One starwars E2E test fails — pre-existing, not caused by this PR.** `com.example.starwars.service.test.BatchResolverIntegrationTest.batch resolver efficiently resolves film characters` fails with `io.micronaut.http.client.exceptions.ReadTimeoutException`. I reproduced the same failure on plain `origin/main` (commit `d9bf5b18`, no slice-1 changes applied), with the same test, same exception. The other 3 tests in the same suite pass; the rest of `:starwars:test` passes.

Surfacing per the rebase-complexity rule rather than attempting to fix in this PR.

---

_Above is generated by Claude Code._

gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt
```diff
@@ -36,37 +39,60 @@ open class ViaductApplicationExtension(objects: ObjectFactory) {
...
+     * [GradleException].
      */
-    fun declaredScopedSchema(id: String, scopes: Set<String>) {
-        val existing = scopedSchemasProperty.getOrElse(emptyMap())
-        if (id in existing) {
+    fun declaredScopedSchemas(vararg entries: Pair<String, Set<String>>) {
+        if (scopedSchemasDeclared) {
+            throw GradleException(
+                "declaredScopedSchemas may only be called once. " +
+                    "Compose convention-plugin contributions into a single varargs invocation.",
```
rstata, line 77: what are convention-plugin contributions?  are you thinking that an application might want to use multiple convention plugins, each defining some scopes and scoped-schemas, to collectively define the full set of scopes and scoped-schemas?

rstata:
@geovannefduarte  - this looks good to me, try bringing it back into treehouse again

xyu307:
**Note for the record:** this PR was closed without being merged through GitHub's merge UI. The work was applied to `main` via a custom commit process as a single squashed commit, [`0835e26e`](https://github.com/airbnb/viaduct/commit/0835e26e63d9403012e38a603754717da05f598b) (`feat(scope): introduce SchemaScoping data class and DSL accumulation surface (#367)`), with the original authorship preserved.

The landed commit's content matches this PR's HEAD verbatim, plus clean conflict resolution against two unrelated upstream commits that landed in the same window:

- `345bc7b9 chore: remove dead grtPackageName configuration from ViaductApplicationExtension` — removed a property and one matching assertion in the test.
- `fa453725 fix: resolve demoapp viaduct deps from fresh mavenLocal` — touched `processResources` configuration in `gradle-plugins/application/build.gradle.kts` and `gradle-plugins/module/build.gradle.kts`.

Linking this commit hash here so the trail from PR ↔ merged code is recoverable later: PR #367 → `0835e26e` on `main`.

---

_Above is generated by Claude Code._

PR #382, opened 2026-06-04
ci: surface CycloneDX SBOMs in the security-scan job summary

geovannefduarte:
## Description

SBOM generation was recently added, but its output was only visible by running the build locally and digging through per-module `build/reports/sbom/` directories — so it was hard to see what our published fat JARs actually bundle.

- The `cve-parity` job now renders each publication's CycloneDX SBOM into the **GitHub Actions job summary**: an overview table plus a collapsible component list (dependency / version / license) per fat JAR. The raw `cyclonedx.json` files are also uploaded in the `security-scan-reports` artifact.
- Also fixes a latent bug: the CycloneDX 1.10.0 plugin is incompatible with Gradle's configuration cache (on by default), under which it **silently emits an empty SBOM** — so the existing fine-grained `trivy sbom` scans had nothing to scan. The task is now marked `notCompatibleWithConfigurationCache`, so SBOMs are populated both locally and in CI.

- Generated SBOMs for all 7 publications (302 components) and verified the formatter output both as text and rendered through GitHub's GFM engine.
- Confirmed the config-cache fix by counterfactual: `0 → 38` components; the build now succeeds (previously failed/empty).
- Added 20 unit + golden tests for the formatter, run by the existing `python-tests` job (`python3 -m unittest discover -s .github/scripts/tests`). `actionlint` clean on the workflow.
- Docs: added a "SBOM in the CI job summary" section to `config/security/README.md`.

image

### DIFF
--- a/.github/scripts/format_sbom_summary.py
+++ b/.github/scripts/format_sbom_summary.py
@@ -27,11 +27,16 @@ Notes:
     object with `id` (SPDX), a license object with `name` (non-SPDX), or an SPDX
     `expression`. All three are handled; the bulky embedded license `text` is
     ignored.
+  - With multiple scan dirs, SBOMs are de-duplicated by file path (so passing a
+    parent dir and a nested module dir doesn't double-count), and a module name
+    found in more than one root is qualified with its scan root so same-named
+    modules stay distinct instead of overwriting each other.
 """
 
 import json
 import os
 import sys
+from collections import Counter
 
 
 def find_sbom_reports(scan_dir):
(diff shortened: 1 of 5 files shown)


rstata:
> • The new SBOM summary script loses data in its documented multi-directory mode because reports are keyed only by module name. That makes the patch incorrect even though the single-directory workflow path appears to work.
>
>  Review comment:
>
>  - [P3] Avoid dropping duplicate module names across scan roots — /home/raymie_stata/repos/viaduct/.github/scripts/format_sbom_summary.py:114-124
>    If this formatter is invoked with more than one scan directory (the CLI explicitly supports <dir> [<dir> ...]), two roots that  both contain a module with the same relative name, such as api, collide in parsed[module]/failed[module]. The later SBOM silently overwrites the earlier one, so the overview totals and per-publication details are wrong as soon as someone summarizes multiple builds or passes both a parent directory and a nested module directory.

rstata (approved):

PR #387, opened 2026-06-08
fix(ci): make demoapp pipeline re-run-safe and reap stale tmp branches

geovannefduarte:
## Description

Re-running the demoapp pipeline can't currently succeed: the in-run `cleanup` deletes the `tmp/<ref>` branches in `viaduct-dev/*`, but "Re-run failed jobs" doesn't re-run `push-demoapps`, so `check-demoapps` has nothing to clone → "remote branch not found" (this is what alerted on [#383](https://github.com/airbnb/viaduct/pull/383)). Separately, ~12 `tmp/*` branches have already leaked from dispatch paths that never clean up (e.g. `e2e-snapshot-test.yml` has no cleanup job).

This makes the pipeline re-run-safe and bounds the leak:

1. **Gate in-run cleanup on success** (`demoapps-nightly-check.yml`): only delete the `tmp/*` branches when `check-demoapps` succeeded. A failed run keeps its branches, so "Re-run failed jobs" can re-clone and pass.
2. **Scheduled sweep** (`sweep-tmp-branches.yml`): daily, deletes `tmp/*` branches whose latest commit is older than 7 days across the 5 `viaduct-dev` repos. Reaps the existing leak and bounds the branches kept by (1). `tmp/*`-only, reusing the validate-before-delete guard from [#386](https://github.com/airbnb/viaduct/pull/386). Manual dispatch previews by default (`dry_run=true`).
3. **Clearer clone failure** (`check-published-demoapps.yml`): if the branch is gone (e.g. re-running a run older than the sweep TTL), print "dispatch a fresh run; re-running won't recreate it" instead of a raw git error.

Degrades gracefully: even if GitHub doesn't re-run the (now-skipped) cleanup on a partial re-run, the scheduled sweep is the backstop, so re-run safety doesn't hinge on exact "Re-run failed jobs" semantics.

- `actionlint` passes on all three workflows (YAML + shellcheck of the `run:` scripts).
- **Sweep logic dry-run against the live repos** (read-only): with TTL=7d it correctly flags all 12 stale branches (26–45 days old, across cli/jetty/micronaut/starwars) as WOULD-DELETE and finds none in ktor-starter. Confirmed the `tmp/*` branches are Copybara commits whose committer date tracks push time (distinct dates per dispatch), so committer date is a sound age signal.

### DIFF
--- a/.github/workflows/sweep-tmp-branches.yml
+++ b/.github/workflows/sweep-tmp-branches.yml
@@ -1,0 +1,100 @@
+name: Sweep Stale tmp/* Branches
+# Reaps leftover tmp/<ref> branches in the viaduct-dev/* demoapp repos.
+#
+# The demoapp pipelines push ephemeral tmp/<ref> branches to those repos. The in-run
+# cleanup in demoapps-nightly-check.yml only deletes them when the run succeeded (so a
+# failed run stays re-runnable), and some dispatch paths (e.g. e2e-snapshot-test.yml)
+# don't clean up at all. This scheduled sweep bounds the resulting leak by deleting
+# tmp/* branches whose latest commit is older than the TTL.
+
+on:
+  schedule:
+    - cron: '30 6 * * 1-5'   # 06:30 UTC weekdays, just after the nightly build
+  workflow_dispatch:
+    inputs:
+      max_age_days:
+        description: "Delete tmp/* branches whose latest commit is older than this many days"
+        required: false
+        default: "7"
+        type: string
+      dry_run:
+        description: "List what would be deleted without deleting (manual runs preview by default)"
+        required: false
+        default: true
+        type: boolean
+
+permissions:
+  contents: read
+
+concurrency:
+  group: sweep-tmp-branches
+  cancel-in-progress: false
+
+jobs:
+  sweep:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/sweep-tmp-branches.yml
```diff
@@ -0,0 +1,100 @@
...
+      matrix:
+        demoapp: [cli-starter, jetty-starter, ktor-starter, micronaut-starter, starwars]
+    name: Sweep ${{ matrix.demoapp }}
+    steps:
+      - name: Delete stale tmp/* branches
+        env:
+          GH_TOKEN: ${{ secrets.VIADUCT_GRAPHQL_GITHUB_ACCESS_TOKEN }}
+          DEMOAPP: ${{ matrix.demoapp }}
+          # On schedule these inputs are empty -> default to a real sweep with a 7-day TTL.
+          MAX_AGE_DAYS: ${{ inputs.max_age_days || '7' }}
```
rstata, line 47: won't this mask the default you attempt to set on line 53?  i'd suggest using the `|| default` here for both parameters and removing lines 53-54

geovannefduarte: Applied this change in treehouse

PR #388, opened 2026-06-08
ci: surface security-scan report in the job summary for fork PRs

geovannefduarte:
## Description

PRs from forks run the Security Scan with a **read-only `GITHUB_TOKEN`** (GitHub's security default, untrusted PR code must not receive a write token or secrets). Two consequences today:

- The CVE-parity report was delivered **only** via the sticky PR comment, which a read-only token can't post, so fork/outside contributors never saw it inline.
- The comment step threw `HttpError: Resource not accessible by integration` and exited 1, leaving two red **annotations** on every fork PR (the run still passed, since the step is `continue-on-error`).

**Key changes**:

1. **`ci`:** the `cve-parity` job also appends `security-scan-report.md` to the **job summary** (`$GITHUB_STEP_SUMMARY`) — token-free, so the report is visible on every run, forks included (matching how the SBOM summary already works).
2. **`fix(ci)`:** the sticky-comment step now runs only on same-repo PRs, and its API calls are wrapped in try/catch — a read-only 403 logs a `core.notice` instead of failing with an unhandled error.

The sticky comment still posts on same-repo PRs; the uploaded artifact is unchanged.

- `actionlint` clean on `security-scan.yml`.
- `node --check` on the extracted github-script body — syntax OK.
- End-to-end will confirm on the next fork PR (report in the job summary, no annotations) and a same-repo PR (sticky comment still posts).
- Workflow-only change; no script/unit-test changes.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -161,7 +161,11 @@ jobs:
       # no peter-evans actions are used here). Finds the existing comment by the
       # marker the script emits and updates it, else creates one.
       - name: Post sticky PR comment
-        if: always() && github.event_name == 'pull_request'
+        # Same-repo PRs only: PRs from forks get a read-only GITHUB_TOKEN that can't
+        # post comments. Those still get the report via the job summary + artifact.
+        if: >-
+          always() && github.event_name == 'pull_request'
+          && github.event.pull_request.head.repo.full_name == github.repository
         continue-on-error: true
         uses: actions/github-script@v8
         with:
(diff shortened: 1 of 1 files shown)


rstata:
github has a 1MB limit for each `GITHUB_STEP_SUMMARY` and errors happen when you exceed that.  so we should put some kind of limit on how much we write out.  frankly even 1MB seems like too much to dump into these summaries, how big is this output typically?

geovannefduarte:
> github has a 1MB limit for each `GITHUB_STEP_SUMMARY` and errors happen when you exceed that. so we should put some kind of limit on how much we write out. frankly even 1MB seems like too much to dump into these summaries, how big is this output typically?

I did a quick sizing pass: with the current 7 publications (~300 dependency rows), the summary is only ~25–30 KB, so we're well below the 1 MiB GitHub limit. We'd need something on the order of 20× the current dependency count to get close. Also, the step and job are continue-on-error, so even if we ever overflowed, it wouldn't block a PR; the full report would still be available in the security-scan-reports artifact. Here's an example: https://github.com/airbnb/viaduct/actions/runs/27167964656

geovannefduarte:
Done in [387251cb](https://github.com/airbnb/viaduct/pull/388/changes/387251cbb000f8ff75f0fe3f8b227b85b22dda7e)
Added a note right at the `$GITHUB_STEP_SUMMARY` write explaining we're intentionally not capping it: GitHub's per-step limit is 1 MiB, this output runs well under ~50 KB as of mid-2026  and we should revisit (truncate + point to the security-scan-reports artifact) if it grows toward the cap

PR #389, opened 2026-06-09
docs(release): add pre-publication SBOM inspection gate

geovannefduarte:
## Description

Our published fat JARs can accidentally bundle test/build-only libraries — e.g. via a stray `testFixtures(...)` dependency that pulls in junit, mockk, bytebuddy, javassist, assertj, and friends. This slips through CI (every test still passes) but ships the bad dependencies to consumers, and the `test-fixtures` jar is especially prone to it. We now publish a CycloneDX SBOM per fat JAR, so this adds a quick inspection step the release manager runs to catch such inclusions **before** anything is published.

This adds a **Pre-publication SBOM inspection** section to the release runbook (`.github/impldocs/release-runbook.md`), referenced as a pre-publish check from both publish points — Step 5 (RC) and Step 8 (final) — with a Quick-Reference row and Overview pointers. It's intentionally lightweight (open and eyeball, not a scripted gate):

1. **Generate** the SBOMs: `./gradlew -p publications cyclonedxBom --no-configuration-cache`.
2. **Open and skim** each one (in an editor / `more`, or a one-liner that lists just the coordinates), with a few real sample lines so you know what a normal entry looks like.
3. **Denylist** — a published runtime jar should contain none of junit, mockk, mockito, byte-buddy, objenesis, javassist, assertj, strikt, hamcrest, truth (or anything with `test`/`assert`/`mock` in the name); a hit means a test dependency leaked, so stop and fix. Includes a one-line caveat that `kotlin`/`kotest`/`opentest4j` appear in the SBOM but are excluded from the packaged jar, so they're expected.

Docs-only; no build or workflow changes.

This change *is* the documentation (a new runbook section). The commands in it were dry-run against the `publications` build before writing them down:

- `cyclonedxBom --no-configuration-cache` produced 7 populated SBOMs. Confirmed the footgun the step calls out: under the default configuration cache `cyclonedxBom` silently emits **empty** SBOMs — which is why the step uses `--no-configuration-cache` and says to re-run if a file comes out nearly empty.
- The coordinate-listing one-liner was run against the real SBOMs; the sample lines shown in the runbook are taken from actual output.
- Verified the "expected, not a leak" caveat: `test-fixtures`' SBOM lists `kotest`/`opentest4j`/`kotlin`, but the packaged jar contains **0** of them (they're `exclude()`d) — so the caveat is accurate and keeps the `test`/`assert`-in-the-name heuristic from false-flagging them.
- Markdown checks: code fences balanced, all anchor links resolve, section renders via GitHub's GFM engine.

### DIFF
--- a/.github/impldocs/release-runbook.md
+++ b/.github/impldocs/release-runbook.md
@@ -7,10 +7,10 @@ The release process follows these steps:
 1. **Bump version** on main to the next SNAPSHOT (e.g., `0.8.0-SNAPSHOT`)
 2. **Create release branch** from the commit before the bump (e.g., `release/v0.7.0` with initial VERSION `0.7.0`)
 3. **Validate** — run CI across all Java/OS combinations
-4. **(Optional) Publish release candidate** — publish `X.Y.Z-rc.N` artifacts, push RC demo apps, and verify them
+4. **(Optional) Publish release candidate** — inspect the published-jar SBOMs, then publish `X.Y.Z-rc.N` artifacts, push RC demo apps, and verify them
 5. **Generate changelog** and review with the team
 6. **Confirm release** at the team meeting
-7. **Publish final release** — one command that publishes `X.Y.Z`, pushes demo apps to `main`, verifies them, creates the tag, and publishes the GitHub release
+7. **Publish final release** — inspect the published-jar SBOMs, then run one command that publishes `X.Y.Z`, pushes demo apps to `main`, verifies them, creates the tag, and publishes the GitHub release
 8. **Verify** — pull latest version of Star Wars demo app and verify it passes its tests
 
 ## Quick Command Reference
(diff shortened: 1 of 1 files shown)


.github/impldocs/release-runbook.md
```diff
@@ -32,6 +32,7 @@ export GH_USER="your-github.com-username"
 | [4. Validate build](#4-validate-build) | `gh workflow run ci-trigger.yml --ref release/v${RELEASE_VER}` |
 | [5. Publish RC](#5-publish-release-candidate-optional) | `./gradlew setVersion -PsetVersion=${RELEASE_VER}-${RC_VER}`, commit, push, `gh workflow run release.yml -f release_version=${RELEASE_VER} -f rc_ver=${RC_VER}` |
 | [6. Changelog](#6-generate-changelog) | `generate_changelog.py origin/release/v${PREV_VER} HEAD` |
+| [SBOM inspection](#pre-publication-sbom-inspection) (before each publish) | `./gradlew -p publications cyclonedxBom --no-configuration-cache`, then inspect |
```
rstata:
```suggestion
| 7. [SBOM inspection](#pre-publication-sbom-inspection) (before each publish) | `./gradlew -p publications cyclonedxBom --no-configuration-cache`, then inspect |
```

geovannefduarte: ✅ Fixed

gokhan-ozgozen (approved):

PR #393, opened 2026-06-12
feat(scope): validate viaductApplication schema-scoping DSL (slice 2 of #361)

### LINKED ISSUE #371: Schema Scoping: Slice 2 — Extension validation (hybrid timing)
This is the 2nd of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** A (build-time) · **Depends on:** Slice 1 (#366) · **Status:** Required

## Goal

Validate the schema-scoping DSL inputs at configuration time and emit clear `GradleException`s for any rule violation, using **hybrid validation timing**:

- **Per-ID syntax checks** (regex on a single ID) fire **at setter time** → precise stack trace at the offending `build.gradle.kts` line.
- **Cross-property invariants** (each scoped schema's scope set ⊆ declared universe; reserved IDs) fire **after the declarations settle** → no dependency on whether `declaredSchemaScopes` or `declaredScopedSchemas` is written first, and all violations are batched into one failure.

This reflects the Slice 1 DSL shape: a single `declaredSchemaScopes(scopes)` and a single `declaredScopedSchemas(vararg "id" to scopes)` (no cross-call accumulation). The **non-empty universe** rule and **intra-call duplicate** rejection ship in Slice 1's setters; this slice adds format, reserved-ID, and subset validation.

## Why this is its own PR

Keeps the rule logic reviewable in isolation, separate from the SDL-level `@scope` validation in Slice 3 and the runtime semantics in Series B.

## Files touched

- `core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScopingValidator.kt` — validation logic, regex constants, error codes/messages. Co-located with `SchemaScoping`.
- `gradle-plugins/common/src/main/kotlin/viaduct/gradle/ViaductApplicationExtension.kt` — DSL methods call per-ID syntax checks immediately.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt` — `afterEvaluate` hook runs the cross-property validator.
- `core/service/api/src/test/kotlin/viaduct/service/api/scoping/SchemaScopingValidatorTest.kt` — unit tests on the pure logic.
- `gradle-plugins/application/src/test/kotlin/viaduct/gradle/ScopeConfigurationValidationTest.kt` — TestKit tests on the diagnostic surface.

### Validator (split by what it inspects)

```kotlin
package viaduct.service.api.scoping

object SchemaScopingValidator {
    // The two ID kinds live in different contexts, so two patterns:
    const val SCOPE_ID_PATTERN = "^[a-z][a-z0-9_]*$"        // scope IDs appear in @scope(to: [...]) — SDL identifier shape
    const val SCHEMA_ID_PATTERN = "^[A-Za-z][A-Za-z0-9_]*$" // scoped-schema IDs are API names callers type (PUBLIC_API, publicApi)
    val RESERVED_SCHEMA_IDS = setOf("FULL", "NONE")

    // Per-ID syntax — fires synchronously at setter time.
    fun validateScopeId(id: String): Result<String> =
        if (Regex(SCOPE_ID_PATTERN).matches(id)) Result.success(id)
        else Result.failure(InvalidScopeIdException(
            code = "SCOPE_ID_FORMAT_INVALID",
            message = "Scope id '$id' does not match $SCOPE_ID_PATTERN. " +
                      "Scope ids appear in @scope(to: [...]) and follow GraphQL identifier conventions."))

    fun validateSchemaId(id: String): Result<String> = when {
        id in RESERVED_SCHEMA_IDS -> Result.failure(InvalidScopeIdException(
            code = "SCHEMA_ID_RESERVED",
            message = "Scoped-schema id '$id' is reserved by Viaduct and cannot be declared."))
        Regex(SCHEMA_ID_PATTERN).matches(id) -> Result.success(id)
        else -> Result.failure(InvalidScopeIdException(
            code = "SCHEMA_ID_FORMAT_INVALID",
            message = "Scoped-schema id '$id' does not match $SCHEMA_ID_PATTERN. Examples: PUBLIC_API, publicApi, FullApi."))
    }

    // Cross-property — fires after both declarations settle.
    fun validate(scoping: SchemaScoping): List<ScopingError> {
        val errors = mutableListOf<ScopingError>()

        // Subset: every scoped schema's scopes must be declared in the universe.
        scoping.scopedSchemas.forEach { (id, scopes) ->
            val unknown = scopes - scoping.scopeUniverse
            if (unknown.isNotEmpty()) errors += ScopingError(
                code = "SCOPE_SET_NOT_SUBSET",
                message = "Scoped schema '$id' references scopes not declared in declaredSchemaScopes: " +
                          "${unknown.sorted()}. Declared scopes: ${scoping.scopeUniverse.sorted()}.")
        }

        // Scoped schemas declared without a universe (only reachable if the Slice 1 non-empty guard is bypassed).
        if (!scoping.isScoped && scoping.scopedSchemas.isNotEmpty()) errors += ScopingError(
            code = "SCOPED_SCHEMAS_WITHOUT_UNIVERSE",
            message = "declaredScopedSchemas is set but no scope universe is declared. " +
                      "Call declaredSchemaScopes(...) with at least one scope, or remove the scoped schemas.")

        return errors
    }
}
```

Per-ID checks throw `InvalidScopeIdException`; cross-property checks return a list and the plugin throws a single aggregating `GradleException`. Error codes + templated, remediation-bearing messages are the first build-time instance of the error-catalogue practice this work seeds.

### Extension — fail-fast at setter time

```kotlin
fun declaredSchemaScopes(scopes: Set<String>) {
    require(scopes.isNotEmpty()) { "declaredSchemaScopes(...) requires at least one scope (a declared universe is non-empty)." }
    scopes.forEach { SchemaScopingValidator.validateScopeId(it).getOrThrow() }
    // … store (single declaration; intra-call duplicate rejection from Slice 1)
}

fun declaredScopedSchemas(vararg schemas: Pair<String, Set<String>>) {
    schemas.forEach { (id, scopes) ->
        SchemaScopingValidator.validateSchemaId(id).getOrThrow()
        scopes.forEach { SchemaScopingValidator.validateScopeId(it).getOrThrow() }
    }
    // … reject duplicate IDs among entries (Slice 1), then store
}
```

### Plugin — cross-property in `afterEvaluate`

```kotlin
project.afterEvaluate {
    val errors = SchemaScopingValidator.validate(appExt.schemaScoping.get())
    if (errors.isNotEmpty()) throw GradleException(
        "viaductApplication scope configuration is invalid:\n" +
            errors.joinToString("\n") { "  - [${it.code}] ${it.message}" })
}
```

## Why hybrid timing

Per-ID syntax at the setter gives a precise stack trace for typos like `setOf("BAD-ID")`. The subset check spans two separate declarations, so running it after configuration settles avoids any dependency on which declaration appears first and lets all violations batch into one readable message.

## Test plan

**Unit** (`SchemaScopingValidatorTest.kt`, no Gradle):
1. `validateScopeId` accepts `public`, `internal_v2`; rejects empty / uppercase / hyphenated / leading-digit.
2. `validateSchemaId` accepts `PUBLIC_API`, `publicApi`, `FullApi`; rejects hyphenated / leading-digit / leading-underscore.
3. `validateSchemaId` rejects reserved `FULL`, `NONE`.
4. `validate(SchemaScoping.EMPTY)` → no errors.
5. `validate` flags subset violations, naming the offending IDs.
6. `validate` flags scoped-schemas-without-universe.
7. `validate` aggregates multiple errors.

**TestKit** (`ScopeConfigurationValidationTest.kt`, ~3 cases):
8. Valid config configures cleanly.
9. Single syntax error surfaces with file:line in `BuildResult.output`.
10. Multiple errors (subset + bad ID) appear together in one failure message.

**Config-cache gate:** the multi-error TestKit case must pass with `--configuration-cache` — the `afterEvaluate` hook reading `schemaScoping.get()` must be cache-safe. **Hard gate before merge.**

## Out of scope (deferred)

- Intra-call duplicate rejection + non-empty guard — Slice 1.
- SDL `@scope` validation against the declared universe — Slice 3.
- Conditional `@scope` directive — Slice 4.

## LOC estimate

~250 LOC: ~120 validator, ~30 setter wiring, ~20 `afterEvaluate` hook, ~80 tests.

## Dependencies

Slice 1 (#366) — the `SchemaScoping` data class + DSL must exist.

xyu307:
Closes #371 (slice 2 sub-issue of the schema-scoping umbrella #361)

## Summary

Implements slice 2 of the schema-scoping rollout (umbrella [#361](https://github.com/airbnb/viaduct/issues/361) — sub-issue [#371](https://github.com/airbnb/viaduct/issues/371)): validates the `viaductApplication` scope declarations at configuration time with **hybrid timing** — per-ID syntax inside the DSL setters (precise file:line stack), cross-property invariants in `afterEvaluate` (no DSL call-order dependency, batched errors). Every diagnostic carries a stable error code rendered as `[CODE] message`.

A companion commit retrofits the five existing slice-1 throws onto the same `[CODE] message` convention.

## What's in the diff

1. **`feat(scope): validate schema-scoping DSL configuration with hybrid timing`** — pure validator + error catalogue in `core/service/api/.../scoping/`, setter wiring in `ViaductApplicationExtension`, `afterEvaluate` hook in `ViaductApplicationPlugin`, unit + TestKit coverage.
2. **`feat(scope): tag slice-1 DSL diagnostics with stable error codes`** — `[CODE]` prefix on the five existing throws. No behavior change beyond message text.

## Reviewer decisions honored

Per the clarification thread on the issue: plain `GradleException` with `[CODE]` prefix, no subclass hierarchy; `SCOPED_SCHEMAS_WITHOUT_UNIVERSE` fires even for `"X" to emptySet()`; universe-only is valid; DSL rejects `*` (SDL handling is slice 3); reserved IDs `FULL`/`NONE`.

## Things to flag for reviewer attention

Three observable behaviors that aren't bugs but deserve a sanity-check before merge:

**a. Validation timing — which DSL patterns are covered.** Per-ID checks fire inside the setters and are always active. Cross-property checks fire once in `project.afterEvaluate` registered during plugin apply. Synchronous DSL calls from the top-level build script, from `subprojects { }`, or from a convention plugin's `apply()` are all fully validated. DSL contributions deferred to `project.afterEvaluate { }`, `gradle.projectsEvaluated { }`, or any later-than-end-of-evaluation hook bypass the cross-property step (per-ID checks still fire). Documented this rather than generalizing the hook (e.g. via `gradle.taskGraph.whenReady`) because deferred-contribution convention plugins aren't the established pattern in this codebase — happy to swap the hook if you'd prefer the broader coverage.

**b. Reserved-id check is case-sensitive.** `FULL` and `NONE` are reserved, but `Full`, `full`, `None`, `none` pass as ordinary IDs because the check is `id in setOf("FULL", "NONE")`. Intentional — uppercase sentinels are tradition — but a one-line `equals(ignoreCase = true)` tightening if you prefer.

**c. Error code naming deviation from the spec.** The spec calls the subset-violation code `SCOPE_SET_NOT_SUBSET`. The implementation lands as `SCOPED_SCHEMA_UNKNOWN_SCOPE` because it describes the offending side (a scoped-schema entry references an unknown scope) rather than the algebraic relation, and reads more naturally in the error message body. Happy to revert to the spec name.

## Test plan

- [x] `:core:service:api:test :gradle-plugins:common:test :gradle-plugins:application:test :core:service:api:apiCheck` all green.
- [x] TestKit happy path under `--configuration-cache --configuration-cache-problems=fail` — `TaskOutcome.SUCCESS` on `:help`, no cache problems.
- [x] TestKit failing paths (per-ID + cross-property aggregation) surface the right codes and stack frames; cross-property case also runs under `--configuration-cache --configuration-cache-problems=fail`.

_Above is generated by Claude Code._


core/service/api/src/test/kotlin/viaduct/service/api/scoping/SchemaScopingValidatorTest.kt
```diff
@@ -0,0 +1,147 @@
...
+
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+import org.junit.jupiter.api.Test
+import viaduct.apiannotations.ExperimentalApi
+
+@OptIn(ExperimentalApi::class)
+class SchemaScopingValidatorTest {
```
geovannefduarte, line 11: Please, add a positive case for the full-alias-with-universe pattern, since it's the one valid path not yet pinned and it's a documented usage

xyu307: Good catch — the empty-set-as-full-alias path with a universe declared was unpinned.

Added in 8936476c as two tests next to the existing universe-only positive case:
- `validate accepts a sole full-alias entry when a universe is declared` — minimal positive case for the documented `"FULL_ALIAS" to emptySet()` usage.
- `validate accepts a full-alias entry alongside a normal scoped schema` — pins that the empty-set entry doesn't taint validation of a normal sibling (the subset-check loop's empty-unknown short-circuit must not interfere with the next entry).

Split into two so a regression that breaks only one interaction can't hide behind the other. Comments in the test bodies name the spec contract (full-schema alias from `decisions.md` / `implementation-plan.md`) and contrast with the no-universe rejection case below, so future readers see the distinction without grepping the spec.

No production code change — the validator already handles this correctly at `SchemaScopingValidator.kt:100-108` (`emptySet() - universe = emptySet()` short-circuits the unknown-scope check). This commit just pins the behavior so future refactors can't regress it silently.

xyu307: Update following the rebase onto current `main` — superseding my earlier reply that referenced the pre-rebase SHA. The full-alias coverage is now in commit `fc28b62a` (instead of `8936476c`); same two tests, same shape:

- `validate accepts a sole full-alias entry when a universe is declared` — minimal positive case for the documented `"FULL_ALIAS" to emptySet()` usage.
- `validate accepts a full-alias entry alongside a normal scoped schema` — pins that the empty-set entry doesn't taint validation of a normal sibling (the subset-check loop's empty-unknown short-circuit must not interfere with the next entry).

Split into two so a regression that breaks only one interaction can't hide behind the other. Comments in the test bodies name the spec contract (full-schema alias from `decisions.md` / `implementation-plan.md`) and contrast with the no-universe rejection case below, so future readers see the distinction without grepping the spec.

No production code change for this comment — the validator already handles this correctly at `SchemaScopingValidator.kt:100-108` (`emptySet() - universe = emptySet()` short-circuits the unknown-scope check). The added tests pin the behavior so future refactors can't regress it silently.

---

While picking this up I also rebased the branch onto current `main`, which brought in two upstream changes that produced clean textual merges but semantic conflicts. Folded the resolutions into the existing commits to keep the series at three; calling them out here for visibility:

1. **`4e59880f` (`build!: stop publishing shared thin JARs ...`):** the direction is right — keeping `viaduct.service.api.*` types out of the plugin JARs' visible surface avoids accidental coupling. The narrowing of `schemaScoping` to `internal` and the `api` → `implementation` demotion on the `:common` → `service:api` edge were both consistent with that surface story. The piece this slice surfaces is that `schemaScoping` does have an in-tree cross-module reader: the `afterEvaluate` validation hook this PR adds in `:application` calls `appExt.schemaScoping.get()` from a different Kotlin module than where it's defined (Kotlin `internal` is module-scoped). Slice 3 will follow with `Property<SchemaScoping>` as a typed `AssembleCentralSchemaTask` input, which has the same cross-module need. Resolved by re-promoting `schemaScoping` to public `val` but annotating it `@InternalApi`, so BCV's `nonPublicMarkers` config still drops it from the published ABI listing — the spirit of `4e59880f` (no `service.api` symbols visible to plugin consumers) is preserved; the symbol stays reachable inside the build. `:common`'s `service.api` dep is re-promoted to `api(...)` for the same reason (callers in `:application` need to resolve `SchemaScoping` / `SchemaScopingValidator` at compile time).

2. **`a41b57d5` (`refactor(viaduct-oss): migrate test assertions off assertj-core, strikt-core, kotlin-test, ...`):** both test files this slice introduces were authored against `kotlin.test.*` (the pre-`a41b57d5` convention). Migrated them to the new convention per the same commit's template — JUnit 5 `Assertions.*` for boolean/equality and Kotest `shouldContain` for substring matches. Two JUnit-5 specifics worth noting: `assertNotNull` doesn't Kotlin-smart-cast, so call sites use `err!!.code`; and `assertEquals(emptyList(), ...)` needs an explicit type witness since JUnit's signature is `(Object, Object)`.

Local build is green: `:core:service:api:test`, `:gradle-plugins:application:check`, `:gradle-plugins:common:check` all pass; `apiCheck` is UP-TO-DATE so there's no BCV public-surface drift.

rstata:
overall this PR looks good, however i don't like the "hybrid validation," which i think we can eliminate by tweaking the API for configuring schema, see:

https://gist.github.com/rstata/d736a721a7559f9610944a72ceac2712

xyu307:
## DSL pivot pushed — single `declareScoping { ... }` block

Commit [`1ec158ac`](https://github.com/airbnb/viaduct/pull/393/commits/1ec158ac) implements the [design update on #371](https://github.com/airbnb/viaduct/issues/371#issuecomment-4847635792) and folds in the code-review response for this PR. Highlights:

- **DSL surface.** `declaredSchemaScopes` + `declaredScopedSchemas` + `afterEvaluate` replaced with a single `declareScoping { scopes(...); scopedSchema(...) }` receiver block. Per-ID syntax and per-call invariants throw synchronously inside the offending DSL call; cross-property checks run at end-of-block via `SchemaScopingBuilder.build()`. Failures land at the user's build-script line, not in a deferred callback.
- **New error code.** `SCHEMA_SCOPE_DUPLICATE_ID` covers duplicates inside one `scopes(...)` vararg list. Codes total 10; every code is now exercised by at least one test.
- **Ordering fix in `scopedSchema(...)`.** Scope-ID shape validation now runs before the duplicate-schema-ID guard, so a copy-pasted line with a typo'd scope surfaces `SCOPE_ID_FORMAT_INVALID` at the offending line instead of hiding behind `SCOPED_SCHEMA_DUPLICATE_ID`. Regression test added.
- **`@OptIn` cleanup.** `ViaductApplicationExtension` opts in to both `ExperimentalApi` and `InternalApi` so the `SchemaScopingBuilder()` constructor and `.build()` don't emit warnings.
- **Test rebalance.** `ViaductApplicationScopeValidationTest` (TestKit) trimmed to three representative smoke checks — happy-path under `--configuration-cache`, one per-call DSL failure showing the build-script line, and one cross-property aggregation. Per-code coverage lives in `ViaductApplicationExtensionTest` (`ProjectBuilder`). Dropped `--configuration-cache-problems=fail` on the failure-path aggregation case since the failure never reaches cache write.
- **Assertion style.** Switched to Kotest `shouldContain` throughout, so assertion failures print the actual message body.

Test results (all green):

| Suite | Tests |
| --- | --- |
| `ViaductApplicationExtensionTest` | 16 / 0 fail |
| `ViaductApplicationScopeValidationTest` | 3 / 0 fail |
| `SchemaScopingValidatorTest` | (unchanged, green) |

Deliberately **not** changed, per direction on the code-review findings:
- The `scopingDeclared` latching after a throw stays as-is — calling `declareScoping` twice is a DSL rule violation regardless of what happened inside the first call. No try/catch-for-retry pattern is supported.
- No `@DslMarker` added.
- Validator messages may reference DSL-declaration syntax; JSON manifest is only written after all validation passes, so validator vocabulary isn't manifest-facing.
- `ScopingErrorCodes` stays in `core:service:api` (centralized).

rstata (approved):

geovannefduarte (approved):

Issue #377, opened 2026-06-04
Schema Scoping: Slice 8 — JSON manifest writer


geovannefduarte:
This is the 8th of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** B (runtime consumption) · **Depends on:** Slice 1 (#366) · **Status:** Required

## Goal

Emit `META-INF/viaduct/schema-scoping.json` into the application's main resources. **Build-time only** — no runtime consumer yet (that's Slice 9). Serialize from the unified `SchemaScoping` data class so the writer and the Slice 9 reader share a single source of truth (no mirror format, no drift).

## Files touched

- `core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt` — add `toJsonString(scoping)` and `parse(stream/json)` companion methods (parse used by Slice 9; defined together for cohesion).
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/task/WriteViaductSchemaScopingResourceTask.kt` — **new** `@CacheableTask`.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt` — register the task; wire output into `processResources`.
- Tests in `core/service/api/src/test/` (round-trip) and `gradle-plugins/application/src/test/` (task + JAR contents).

### Companion methods on `SchemaScoping`

```kotlin
companion object {
    const val CURRENT_VERSION = "1"
    const val RESOURCE_PATH = "META-INF/viaduct/schema-scoping.json"
    val EMPTY = SchemaScoping(emptySet(), emptyMap())

    /** Deterministic JSON: declaredScopes sorted; declaredScopedSchemas keys + values sorted; 2-space indent, UTF-8, trailing newline. */
    fun toJsonString(scoping: SchemaScoping): String { ... }

    fun parse(json: String): SchemaScoping { ... }       // used by Slice 9
    fun parse(stream: InputStream): SchemaScoping = parse(stream.bufferedReader().readText())
}
```

JSON shape:

```json
{
  "version": "1",
  "declaredScopes": ["admin", "internal", "public"],
  "declaredScopedSchemas": {
    "FULL_ALIAS": [],
    "INTERNAL_API": ["internal", "public"],
    "PUBLIC_API": ["public"]
  }
}
```

[...]

Issue #374, opened 2026-06-04
Schema Scoping: Slice 5 — Scoped-schema materialization & introspection


geovannefduarte:
This is the 5th of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** A (build-time) · **Depends on:** Slice 1 (#366), Slice 3 · **Status:** Required

## Goal

For each **distinct** declared scope set, materialize the scoped schema via `ScopedSchemaBuilder.applyScopes` and execute the GraphQL introspection query against the result. Implements: every declared scope set must produce a valid scoped schema (C.17) and every resulting scoped schema must support introspection (C.18).

## Why this is its own PR

Materialization can be slow for large schemas; isolating it lets reviewers reason about the perf characteristics independently.

## Files touched

- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/task/AssembleCentralSchemaTask.kt` — add the materialization loop after SDL validation; read scope sets from `schemaScoping`.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ScopedSchemaIntrospectionCheck.kt` — **new** helper: materialize one scope set, run introspection.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt` — wiring (the `schemaScoping` task input already added in Slice 3).

### `AssembleCentralSchemaTask`

```kotlin
@TaskAction
fun taskAction() {
    // … existing assembly + SDL + validateCompleteSchema …

    val scoping = schemaScoping.get()
    if (scoping.isScoped) {
        validateScopedSchemas(allSchemaFiles + sdlFile, scoping)
    }
}

private fun validateScopedSchemas(schemaFiles: Collection<File>, scoping: SchemaScoping) {
    val checker = ScopedSchemaIntrospectionCheck(logger, schemaFiles, scoping.scopeUniverse)
    val errors = mutableListOf<String>()

    // Iterate distinct scope sets — multiple IDs may alias the same set.
    for (scopeSet in scoping.scopedSchemas.values.toSet()) {
        if (scopeSet.isEmpty()) continue            // empty set = full-schema alias; covered by primary validation
        try {
            checker.validate(scopeSet)
        } catch (e: SchemaScopeValidationError) {
            errors += "Scope set ${scopeSet.sorted()}: ${e.message}"
[...]

Issue #373, opened 2026-06-04
Schema Scoping: Slice 4 — Conditional @scope directive injection


geovannefduarte:
This is the 4th of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** A (build-time) · **Depends on:** Slice 1 (#366), Slice 3 · **Status:** Required

## Goal

When the application declares no scope universe, `DefaultSchemaFactory` must **not** define the `@scope` directive AND must not decorate Node / PageInfo / root types with `@scope(to: ["*"])`. Realizes the spec's rule: "If an application declares no scopes, the `@scope` directive should not be defined for that application."

## Why this is its own PR

Touches `DefaultSchemaFactory` — a shared library used by every Viaduct app. Splitting it from Slice 3 keeps the SDL-validation change pure.

## Files touched

- `core/shared/graphql/src/main/kotlin/viaduct/graphql/utils/DefaultSchemaFactory.kt` — add an `includeScopeDirective` parameter to `getDefaultSDL` and `addDefaults`; gate the `@scope` directive definition and the synthetic decorations on it.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/task/AssembleCentralSchemaTask.kt` — pass `includeScopeDirective = schemaScoping.get().isScoped`.
- `core/shared/graphql/src/test/kotlin/viaduct/graphql/utils/DefaultSchemaFactoryTest.kt` — tests for both branches.

### `DefaultSchemaFactory`

```kotlin
fun getDefaultSDL(
    includeNodeDefinition: IncludeNodeSchema = IncludeNodeSchema.IfUsed,
    includeNodeQueries: IncludeNodeSchema = IncludeNodeSchema.IfUsed,
    existingSDLFiles: List<File> = emptyList(),
    includePageInfo: Boolean = true,
    includeRootTypes: Boolean = true,
    includeScopeDirective: Boolean = true,  // NEW
): String { ... }
```

Gate these on `includeScopeDirective`:
- `addDefaultDirectives` — skip `DefaultDirective.SCOPE`.
- `addPageInfoType` — skip `pageInfoBuilder.directive(createScopeDirective(listOf("*")))`.
- `addNodeInterface` / `addNodeQueryFields` — skip the scope directive on the Node interface / Query extension.
[...]

Issue #376, opened 2026-06-04
Schema Scoping: Slice 7 — Type-extension scope correctness rules (optional)


geovannefduarte:
This is the 7th of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** A (build-time) · **Depends on:** Slice 3 · **Status:** Optional

## Goal

Port the compositional type-extension scope rules:
- **A.12** — Union extension scopes must equal the `union ∩ member` intersection.
- **A.13** — Interface-implementation extension scopes must be a subset of `object ∩ interface`.
- **A.14** — Interface fields must exist in every scope where the implementation relationship exists.

A.7 is already enforced via `ScopeDirectiveParser` (Slice 3), so this slice focuses on the trickier compositional rules.

## Why this PR is optional

The spec says: "include A.12–A.14 if practical, because they are schema-correctness rules rather than product policy." If review surfaces real cost — porting the equivalent reference logic 1:1 is non-trivial, and rephrasing it for OSS may take significant effort — punt to a follow-up RFC.

## Files touched

- `core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/rules/UnionExtensionScopeRule.kt` — new.
- `core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/rules/InterfaceImplementationScopeRule.kt` — new.
- `core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/rules/DefaultSchemaValidator.kt` — register the new rules (only when `validScopes` is supplied).
- Tests for each rule.

### `UnionExtensionScopeRule`
For each `extend union U @scope(to: [...])`: compute `U.scopes ∩ M.scopes` for each member `M`; assert the extension's scope set equals that intersection.

### `InterfaceImplementationScopeRule`
For each `extend type O implements I @scope(to: [...])`: compute `O.scopes ∩ I.scopes`; assert the extension's scope set is a subset.

### A.14 (per-scope traversal)
For each `(object O, interface I)` where `O implements I`, for each scope `s` where the implementation relationship is in scope, for each field `f` required by `I`: assert `O.f` exists and has scope `s`. Cost is `O(types × scopes × fields)` — manageable for a real schema, but worth measuring.

## Test plan

- Per-rule unit tests with hand-built schemas: happy path + each violation form.
[...]

PR #394, opened 2026-06-12
feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)

### LINKED ISSUE #372: Schema Scoping: Slice 3 — SDL validation against the declared universe (primary phase)
This is the 3rd of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** A (build-time) · **Depends on:** Slice 1 (#366), Slice 2 · **Status:** Required

## Goal

Add an explicit primary-phase `ScopeUsageRule` to `DefaultSchemaValidator` that runs once per build against the full SDL, taking the declared scope universe as input. **This is the upstream net** — it closes configurations where scope-rule firing today depends on `FilterChildrenVisitor`'s materialization-time traversal.

Implements the PRD-recommended subset for OSS v1: rules **A.1, A.2, A.3, A.4, A.5, A.6, A.7, A.9** explicitly (A.8 is enforced separately; A.10 is extraction behavior; A.11 is covered by A.7). A.8 requires the literal `["*"]` form.

## The coverage-gap problem (why this PR is necessary)

Today the scope-rule logic lives in `ScopeDirectiveParser.metadataForElement(...)`, invoked by `FilterChildrenVisitor` during scoped-schema **materialization**. So scope-rule firing depends on the visitor stack including `FilterChildrenVisitor`, which in turn requires at least one declared scope set to be a **proper subset** of the universe. **Three configurations bypass A.1–A.7 entirely:**

| Configuration | Today's behavior |
|---|---|
| No scopes declared (`isScoped == false`) | `runScopePipeline` doesn't run — no validation |
| Universe declared, but no scoped schemas | Materialization loop iterates zero times — no visitor stack constructed |
| Every declared scope set equals the universe | Only `ValidateRequiredScopesVisitor` (A.9) runs; A.1–A.8 skipped |

A typo `@scope(to: ["typoed"])` in any of these lives in SDL indefinitely and surfaces only if someone later declares a proper-subset scope set. A real correctness bug surfaced during review.

## Files touched

- `core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/rules/ScopeUsageRule.kt` — **new** primary-phase rule (lives with the other schema-validation rules, not with the `SchemaScoping` data class).
- `core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/rules/DefaultSchemaValidator.kt` — register `ScopeUsageRule` when `validScopes` is supplied.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductSchemaValidator.kt` — accept `validScopes: Set<String>?`.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/task/AssembleCentralSchemaTask.kt` — accept `schemaScoping: Property<SchemaScoping>` as `@Input`; pass `scopeUniverse` to the validator.
- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt` — wire the extension's `schemaScoping` to the task input.
- (small) A.8 literal-`["*"]` adjustment in the existing scope visitors.

### New `ScopeUsageRule`

A `ValidationRule` that walks the SDL and enforces:

| Rule | Check |
|---|---|
| **A.1** Invalid scope names | Every string in `@scope(to: [...])` is in `validScopes` or is `*` |
| **A.2** Duplicate scope names | Same scope ID listed more than once in a single `@scope` |
| **A.3** `*` not mixed | `@scope(to: ["*", "public"])` rejected |
| **A.4** Missing `@scope` on participating kinds | Object / input / interface / enum / union types must carry `@scope` when scopes are declared |
| **A.5** Referenced types declare scopes | If A references scope-able type B, B must have `@scope` |
| **A.6** Referencing/referenced share scope | If A references B, `A.scopes ∩ B.scopes ≠ ∅` |
| **A.7** Extension scopes ⊆ base | `extend type T @scope(...)` may not introduce scopes beyond T's base |
| **A.9** Root types have valid `@scope` | `Query` / `Mutation` (and `Subscription` if present) must carry valid `@scope` |

A.8 (directive-retained type uses literal `["*"]`) lives in `ValidateScopesVisitor`; adjust it to require `to == listOf("*")` rather than equality with the universe (more robust). A.10 is `TypeRemovalVisitor` behavior, not an independent rule.

### Validator signature + task wiring

```kotlin
// ViaductSchemaValidator
fun validateSchema(
    schemaFiles: Collection<File>,
    excludeFromViaductValidation: Collection<File> = emptyList(),
    validScopes: Set<String>? = null,  // NEW — non-null ⇒ ScopeUsageRule(validScopes) added to the rule set
): List<GraphQLError>

// AssembleCentralSchemaTask
@get:Input
abstract val schemaScoping: Property<SchemaScoping>

@TaskAction
fun execute() {
    val scoping = schemaScoping.get()
    validateCompleteSchema(
        schemaFiles = allSchemaFiles + sdlFile,
        excludeFromViaductValidation = listOf(sdlFile),
        validScopes = if (scoping.isScoped) scoping.scopeUniverse else null,
    )
}
```

### Relationship to the materialization-time visitors

`ScopeUsageRule` does **not** replace `ScopeDirectiveParser` / `ValidateScopesVisitor` — those still run per scope set during materialization (Slice 5) and catch context-specific cross-references. The split:

- **Primary phase** (this slice) — runs once across the full SDL; closes the three coverage-gap configurations.
- **Materialization phase** (Slice 5) — runs per scope set; catches issues that depend on the materialized state.

## Test plan

Coverage tests for the three previously-bypassed configurations:

1. Universe declared, no scoped schemas, SDL has typo'd `@scope` → **must fail** (passes silently today).
2. Universe declared, every scope set equals the universe, typo'd `@scope` → **must fail** (passes silently today).
3. Universe declared, proper-subset scope set, typo'd `@scope` → must fail (already fails today; keep working).
4. No scopes declared, SDL uses `@scope` → fails because the directive isn't defined (Slice 4).

Plus rule coverage: A.1 undeclared scope, A.2 duplicate, A.3 mixed `*`, A.4 missing `@scope` on object, A.5 referenced type missing `@scope`, A.6 no shared scope, A.7 extension exceeds base, A.8 non-literal `*`, A.9 root without `@scope`, and `@scope(to: ["*"])` always allowed. Unit tests for `ScopeUsageRule` directly in `core/shared/viaductschema/src/test/`.

## Out of scope (deferred)

- Conditional `@scope` directive definition — Slice 4 (for this slice, the directive is still always defined; we only validate references against the universe).
- Per-scope-set materialization & introspection — Slice 5.
- Type-extension correctness rules A.12–A.14 — Slice 7.

## LOC estimate

~250 LOC: ~120 in `ScopeUsageRule.kt`, ~30 validator, ~10 task wiring, ~10 A.8 adjustment, ~80 tests.

## Dependencies

Slice 1 (`SchemaScoping` + `schemaScoping` Provider), Slice 2 (config-cache hardening means `schemaScoping` survives the cache boundary).

xyu307:
Closes #372 (slice 3 sub-issue of the schema-scoping umbrella #361)

### Stacking note (drop after #393 merges and this branch rebases to `main`)

This PR opens against `main` for visibility, but the diff currently includes commits from **#393 (slice 2)** which has not merged yet. **Reviewers: slice 3 itself spans four commits on this branch — `38155950` (initial), `8a238b9d` (review remediation), `d3b65cd5` (framework-offload refactor), and `4d0351bf` (task-input widening).** See the "Commit history" section below for what each commit changed and why. Once #393 merges and this branch rebases to `main`, the diff will collapse to those four commits and this note will be removed.

### Summary

Implements slice 3 of the schema-scoping rollout (umbrella #361 — sub-issue #372): adds an SDL-level `ScopeUsageRule` that enforces `@scope(to: [...])` correctness at compile time against the universe the application declares via `viaductApplication.declaredSchemaScopes(...)`. Every diagnostic carries a stable `SCOPE_*` error code rendered in the build output.

### What's in the diff

- **`ScopeUsageRule`** (new) — `ValidationRule` covering spec items A.1–A.7 and A.9. Visits Object/Interface/Union/Input/Enum + Field defs, parses each `@scope` AppliedDirective, and reports presence/subset/share errors at the offending source location. A.8 (literal-`["*"]` behavior change) is intentionally deferred to a follow-up slice.
- **`ValidationErrorCodes`** — adds 10 new `SCOPE_*` codes (8 from spec items A.1–A.7/A.9 + 1 hardening code `SCOPE_NAME_NOT_VALID_STRING` + 1 legacy-parity code `SCOPE_DIRECTIVE_REPEATED` — see "Decisions called out" below).
- **`DefaultSchemaValidator.create(validScopes)`** — opt-in overload that appends `ScopeUsageRule(validScopes)` to the standard rule list. Null/empty universe returns the existing default singleton, so non-opt-in callers are unchanged.
- **`ViaductSchemaValidator.validateSchema(...)`** — accepts the universe via an optional `validScopes: Set<String>? = null` parameter.
- **`AssembleCentralSchemaTask.schemaScoping: Property<SchemaScoping>`** — `@Input` carrying the full `SchemaScoping` data class (matches the shape called out in #372). Slice 3 only reads `schemaScoping.get().scopeUniverse` to feed `ScopeUsageRule`; the full shape is on the task input because downstream slices (5 reads `scopedSchemas` for materialization; 8 serializes both fields to the JSON manifest) consume additional fields from the same task surface. Wired from `appExt.schemaScoping` in `ViaductApplicationPlugin`.
- **`ValidationContext.computeIfAbsent(rule, key, compute)`** — per-context, per-(rule, key) memoization primitive added so `ScopeUsageRule` can cache per-type scope parsing through the framework instead of a hand-rolled `WeakHashMap` field. Non-breaking for the 12 other rules.
- Tests: 35 unit cases in `ScopeUsageRuleTest`; 3 TestKit cases in `ViaductApplicationSchemaScopeRuleTest` (typo→fail with code, clean→success under `--configuration-cache --configuration-cache-problems=fail`, opt-out→success).

### Commit history (slice 3 commits, oldest first)

The four-commit chain reflects two rounds of post-review remediation plus one alignment edit driven by the slice-4/5/8 cross-slice survey. Detailed rationale lives in each commit message; the summary below points reviewers at what changed and why.

- **`38155950` — `feat(scope): enforce @scope universe at SDL compile time`** — initial implementation of `ScopeUsageRule`, the `ValidationErrorCodes` additions, the `DefaultSchemaValidator.create(validScopes)` overload, and the Gradle wiring. Shipped the task input as `scopeUniverse: SetProperty<String>` (narrower than #372's `Property<SchemaScoping>`) under a "minimal configuration-cache footprint" rationale — see commit `4d0351bf` below for why that was later reverted.

- **`8a238b9d` — `fix(scope): address review findings on ScopeUsageRule`** — Round 2 remediation. Seven defects surfaced by the first review pass, each pinned by a failing-then-passing test: stub-typedef name-collision against `ViaductSchema.VIADUCT_IGNORE_SYMBOL`; `visitField` walking `field.args` so A.5/A.6 fire on argument input types; undeclared-name filter before A.6 shares-scope comparison; new `SCOPE_TO_ARG_NOT_LIST` code for `@scope(to: <non-list>)`; raw-vs-filtered split on `ParsedExtension` so A.7 and the `["*"]` framework-injection carve-out operate on raw names; presence check (A.4 / A.9) keyed on `baseHasScope` (mirrors legacy `_metadataForScopeAllowedElementHelper`); per-`(schema, def.name)` memoization of the scope-name union. See commit body for per-defect detail and the 7 added test cases.

- **`d3b65cd5` — `refactor(scope): offload ScopeUsageRule work to framework + tighten contract`** — Round 3 refactor in response to a second review pass. Seven of ten findings shared one root cause (the rule was duplicating work the validation framework already does, one layer up): added `ValidationContext.computeIfAbsent(rule, key, compute)` and dropped the hand-rolled `WeakHashMap` cache; collapsed `parseAllExtensions` + `computeUnionAllScopeNames` into a single `parseScopes(def)` pass; overrode `visitFieldArg(ctx, arg)` so the framework's per-arg dispatch drives the walk (removing the manual `for (arg in field.args)` loop); dropped a redundant `container` parameter from `checkReferenced`; removed two dead fields. Replaced a silent `if (toArg == null) continue` with a non-null assertion + inline comment documenting that the decoder rejects this upstream (`GJSchemaRawKt.toAppliedDirective` throws on the missing required arg before any rule runs). Added one new test (`directive-structure diagnostics carry extension sourceLocation for IDE jumpability`) as a regression guard on the shared `.withSourceLocation(ext.sourceLocation)` call site, using a backing resource file because the inline-SDL overload of `fromTypeDefinitionRegistry` does not propagate a sourceName.

- **`4d0351bf` — `refactor(scope): widen AssembleCentralSchemaTask input to Property<SchemaScoping>`** — Round 4 alignment with the #372 spec, motivated by the cross-slice survey: slice 5 (#374) reads `scoping.scopedSchemas.values.toSet()` from this same task in the materialization loop and slice 8 (#377) serializes both `declaredScopes` and `declaredScopedSchemas` to the JSON manifest, so the Round 1 narrowed input would have forced an in-flight task-input migration mid-stack. Slice 4 (#373) also wires `includeScopeDirective = schemaScoping.get().isScoped` cleanly against the full shape. Behavior is byte-identical for slice 3 because `schemaScoping.get().scopeUniverse` derives from the same `scopeUniverseProperty` the prior wiring read directly. Config-cache contract preserved by the existing `--configuration-cache-problems=fail` TestKit case running against the new input shape.

### Framework-injection refinement on A.7

`DefaultSchemaFactory` injects `extend type Query @scope(to: ["*"]) { node, nodes }` and `interface Node @scope(to: ["*"])` regardless of the user's narrowed base. The rule treats an extension whose scope set is exactly `{"*"}` as "inherit base, do not broaden" rather than a literal "all scopes" claim — so the framework's injection never trips A.7 against a narrowed base. Pinned by a regression test (`A_7 extension exactly equal to star is treated as inherit base (framework injection contract)`). If the framework's injection convention changes, that test will need updating too.

### Decisions called out (deliberate divergences from legacy `ScopeDirectiveParser`)

An independent audit against the legacy parser surfaced two observable divergences. Both are resolved in this PR — flagged here so the choices are reviewer-visible rather than buried in the diff.

**a. Repeated `@scope` on a single definition is now reported as `SCOPE_DIRECTIVE_REPEATED` (mirrors legacy).** Legacy `ScopeDirectiveParser` threw an uncoded `SchemaScopeValidationError("The scopes directive should not be repeated. Found multiple instances on node.", astNode)` when it saw `type X @scope(...) @scope(...)`. The new rule mirrors that reject semantics under a stable code so build output is greppable and reviewers can pattern-match. The check is per-extension AST node — repeated `@scope` across a base + its extensions is still allowed (governed by A.7). After reporting the error, the rule still combines the scope names across the repeated applications so other A.x diagnostics on the same extension fire in the same build (no silent suppression). Pinned by the regression test `repeated @scope on a single definition is reported as SCOPE_DIRECTIVE_REPEATED (mirrors legacy)` in `ScopeUsageRuleTest`, which cites the legacy reject message in an inline comment.

**b. Non-string elements in `@scope(to: [...])` now report a dedicated `SCOPE_NAME_NOT_VALID_STRING` code instead of being silently dropped.** Legacy `ScopeDirectiveParser` threw `"'$it' is not a StringValue"`. The previous version of this PR used `mapNotNull { (it as? StringLiteral)?.value }`, which silently dropped non-string elements — a latent silent-drop reachable today only through `GJSchemaRaw` (parsed-but-not-semantically-validated), since the directive sig `to: [String!]!` would normally cause graphql-java's semantic validation to reject. Resolution: added the dedicated error code `SCOPE_NAME_NOT_VALID_STRING` rather than piggybacking on A.1's `SCOPE_NAME_NOT_DECLARED` — the defect class is "literal-is-not-a-string", not "name-not-in-the-universe", so the diagnostic accurately reflects the cause. Replaced `mapNotNull` in `parseAllExtensions` with explicit per-element iteration that reports each non-string element at the offending source location. Pinned by the regression test `non-string element in @scope to-list is reported as SCOPE_NAME_NOT_VALID_STRING` in `ScopeUsageRuleTest`.

### Reviewer decisions honored (carried forward from slice 2 and the slice-3 spec review)

- DSL universe is the single source of `validScopes` — pulled via `schemaScoping.get().scopeUniverse` inside `taskAction()` after the task input is bound from `appExt.schemaScoping` (the extension's canonical `Provider<SchemaScoping>`); not duplicated into the task config.
- Plain `SchemaValidationError` shape used by every other rule; no new exception types.
- Wildcard semantics: `"*"` on a base means "all scopes"; on an extension means "inherit base" (no broadening). Framework-injection contract pinned by test.
- Opt-out path preserved: empty/absent universe returns the pre-existing default validator instance.

### Test plan

- [x] `:core:shared:viaductschema:test :core:service:api:test :gradle-plugins:application:test :core:service:api:apiCheck` all green.
- [x] 35-case unit suite over a hand-built `ViaductSchema`, covering every A.x rule + the framework-injection refinement + spec-gap regressions + the two audit-driven legacy-parity pins (`SCOPE_DIRECTIVE_REPEATED` on repeated `@scope`, `SCOPE_NAME_NOT_VALID_STRING` on non-string element) + the seven Round-2 defect pins + the Round-3 sourceLocation regression guard.
- [x] TestKit failing path: typo in `@scope(to: [...])` → `buildAndFail()`, output contains `[SCOPE_NAME_NOT_DECLARED]`, the offending name, and the declared universe.
- [x] TestKit happy path under `--configuration-cache --configuration-cache-problems=fail` — `TaskOutcome.SUCCESS` on `:assembleViaductCentralSchema`, no cache problems. Re-run after `4d0351bf` against the widened `Property<SchemaScoping>` input shape.
- [x] TestKit opt-out: no `declaredSchemaScopes(...)` call → assemble still succeeds under config cache (rule bypassed).


xyu307:
### Review remediation

A code-review pass surfaced 10 findings against `ScopeUsageRule`. Seven are real defects in the new rule; two are framed as design limitations and added to the PR description; one is an efficiency improvement. Each remediated defect is pinned by a failing-→-passing test. Unit suite grows from 27 → 34 cases, still under `./gradlew :core:shared:viaductschema:test`.

| # | What was wrong | Fix |
|---|---------------|-----|
| 1 | Local `VIADUCT_IGNORE_SYMBOL` const held `"VIADUCT_IGNORE_SYMBOL"` (the field NAME) instead of `"VIADUCT_IGNORE"` (the canonical value at `ViaductSchema.kt:647`). The stub typedef tripped A.4 in any path that didn't strip it upstream. | Drop the local const; reference `ViaductSchema.VIADUCT_IGNORE_SYMBOL` directly. |
| 2 | `visitField` only inspected the field's return type; argument types silently bypassed A.5/A.6. | Extract `checkReferenced(...)` and call it once per arg via `field.args`. Label is `Container.field($argName)`. |
| 3 | `unionAllScopeNames` returned raw names, so `sharesScope` matched on undeclared names (`BOGUS == BOGUS` → false-positive share, A.6 silent). | Filter undeclared names inside `computeUnionAllScopeNames` before they reach `sharesScope`. |
| 4 | `@scope(to: "public")` (non-list value) was silently dropped because `as? ListLiteral ?: continue` swallowed it. Reachable today only via `GJSchemaRaw` (parsed-but-not-semantically-validated). | New error code `SCOPE_TO_ARG_NOT_LIST`; report at the offending source location. |
| 5 | A.7 short-circuited on `if (baseScopes.isEmpty()) return` against the filtered set, so a base of `["BOGUS"]` silently passed any concrete narrowing on an extension. | `ParsedExtension` now tracks both `rawScopes` and `scopes`; subset compare uses `rawScopes`. |
| 6 | Framework-injection carve-out `if (other.scopes == setOf("*")) continue` operated on the filtered set, so `["*", "BOGUS"]` collapsed to `{"*"}` and snuck past A.7. | Carve-out now matches `other.rawScopes == setOf("*")` — only the literal `["*"]` shape qualifies. |
| 7 | `checkPresenceOnTypeDef` short-circuited on `anyExtensionHasScope`, letting a base without `@scope` silently pass A.4 whenever any extension carried the directive. Legacy `_metadataForScopeAllowedElementHelper` + `ScopeDirectiveParser.getScopesFromDirective:236` required the BASE AST node to carry `@scope`. | Added `ParsedScopes.baseHasScope`; presence check uses it instead. |
| 10 | `unionAllScopeNames` recomputed across thousands of `visitField` calls. | Memoize per-type via `WeakHashMap<ViaductSchema, MutableMap<String, Set<String>?>>` keyed by `(schema identity, def.name)`. |

#### Known limitations (PR description)

- **#8 — `DefaultSchemaValidator.create()` no-arg bypass.** Build-time scope validation is only enforced via the Gradle `AssembleCentralSchemaTask` wiring. Non-Gradle callers (CLI tooling, demoapps test harness, future convention plugins, IDE inspections) that construct a validator via the no-arg factory will not get `ScopeUsageRule` — they are expected to use the Gradle path. Tracked as a design boundary, not a defect.
- **#9 — `AssembleCentralSchemaTask.scopeUniverse` has no convention.** The Provider chain is bound by `ViaductApplicationPlugin.setupAssembleCentralSchemaTask`. The no-convention shape is intentional under the assumption that the property is always wired by the plugin. Tracked as a design boundary.

#### Legacy cross-check (per finding)

Each defect was cross-referenced against the legacy validation path (`ScopeDirectiveParser`, `_metadataForScopeAllowedElementHelper`, `FilterChildrenVisitor`, `ValidateScopesVisitor`, `ValidateRequiredScopesVisitor`) and against the legacy test files (`ValidateScopesVisitorTest`, `ValidateRequiredScopesVisitorTest`):

- Findings #1, #2, #7 are real semantic divergences from legacy. #1 is unreachable in legacy because materialization strips the stub; #2 is implicitly covered by legacy's `SchemaTraverser`-based walk over every element; #7 mirrors legacy's per-AST-node base check. None had a dedicated legacy test — the new tests pin the parity.
- Findings #3, #5, #6 are defects that legacy avoids by failing earlier (A.1 / A.3 reject undeclared / star-mixed before the subset check runs). The new rule is more permissive (it reports A.1 / A.3 but continues), so these corners are reachable here even though they are not in legacy. No legacy test exists.
- Finding #4 mirrors legacy `ScopeDirectiveParser.kt:256`'s explicit reject under a stable code. No legacy test exists.

#### Validation evidence

- `./gradlew :core:shared:viaductschema:test --tests 'viaduct.graphql.schema.validation.rules.ScopeUsageRuleTest'` — **BUILD SUCCESSFUL**.
- Test report XML: `tests=34, failures=0, errors=0`.
- All 7 new tests are failing-→-passing regressions against the prior commit's `ScopeUsageRule.kt` (each test relies on a defect that the old code exhibited).

xyu307:
## Round 3 — refactor: rule offloads framework work; remaining findings

Pushed `d3b65cd5` to `feat/schema-scoping-slice-3`. Thanks for the second review. The 10 findings sorted into one root cause and three separate items.

### Refactor (covers #1, #3, #4, #5, #6, #9, #10)

Seven of the findings shared a single theme — `ScopeUsageRule` was doing work the validation framework already does, just one layer up. The refactor:

- Adds `ValidationContext.computeIfAbsent(rule, key, compute)` as a per-context, per-(rule, key) memoization primitive (one method, ~10 LOC). Replaces the hand-rolled `WeakHashMap<ViaductSchema, ...>` field in `ScopeUsageRule`. Cache lifetime now equals the validation context — rules no longer own long-lived caches or worry about cross-schema reuse semantics.
- Collapses `parseAllExtensions` + `computeUnionAllScopeNames` into a single `parseScopes(def): ParsedScopes` pass. The derived `union: Set<String>?` view is the only signal downstream `checkReferenced` needs.
- Overrides `visitFieldArg(ctx, arg)` so the framework's existing per-arg dispatch (`ValidationContext.walkRecordFields` lines 113–117) drives field-argument scope checks. Removes the manual `for (arg in field.args)` loop in `visitField`.
- Drops the `container` parameter from `checkReferenced`; both call sites now read `field.containingDef` inside.
- Removes the dead `ParsedExtension.scopes` field and the dead `ParsedScopes.anyExtensionHasScope` property.

### #2 — `if (toArg == null) continue`

Researched and verified: the decoder rejects this upstream. `GJSchemaRawKt.toAppliedDirective` (`core/shared/viaductschema/.../GJSchemaRaw.kt:304`) throws `IllegalStateException("No default value for non-nullable argument to")` before any `ViaductSchema` reaches a validation rule, because `@scope.to` is declared `[String!]!` with no default. A regression test that tried to construct an SDL bypassing it failed at parse time with that exact exception.

Replaced the silent `continue` with `app.arguments[TO_ARG_NAME]!!` and an inline comment documenting the upstream contract. If a future decoder relaxes this check, the fix belongs in the decoder, not in a duplicate rule-layer diagnostic.

### #7 — bare `DefaultSchemaValidator.create()` silently omits `ScopeUsageRule`

**Reviewer attention requested on direction.**

Caller research (`grep -rn 'DefaultSchemaValidator\.create' --include='*.kt'`, excluding `/build/` and `/bin/`):

- `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductSchemaValidator.kt:87` is the only production caller of any `create*` variant. It always passes `validScopes`.
- No production code calls the no-arg `create()`.
- No production code calls `validate(schema)` directly. Only `DefaultSchemaValidatorTest` does (test-only).

So the bare `create()` and `validate(schema)` entry points are test-only surface today but public in shape. A future caller that copy-pastes the no-arg form would get a validator that doesn't enforce schema-scoping rules. Three plausible directions:

1. **Status quo + KDoc.** Document on `create()` that it omits scope rules and is intended for projects that have not opted into scoping.
2. **One-time WARN on `create()`** stating that scope rules are not registered and pointing at `create(validScopes)`. Surfaces the divergence in build logs without changing semantics. Risk: noisy for projects that legitimately have no scope universe.
3. **Make scope rules unconditional** when the schema declares the `@scope` directive at all (detect by directive presence rather than by the universe argument). Strongest guarantee, but couples `DefaultSchemaValidator` to the `@scope` directive name and changes the validation surface for any caller that defines `@scope` without configuring a universe.

This commit does **not** include a behavioral change here. Happy to apply option 1 or 2 in a follow-up.

### #8 — regression guard on `sourceLocation`

Added a test that pins `sourceLocation` attachment on directive-structure diagnostics. The diagnostics emitted from inside the per-extension parse loop (`SCOPE_TO_ARG_NOT_LIST`, `SCOPE_NAME_NOT_VALID_STRING`, `SCOPE_NAME_NOT_DECLARED`, `SCOPE_NAME_DUPLICATE_IN_DIRECTIVE`, `SCOPE_STAR_MIXED_WITH_CONCRETE`, `SCOPE_DIRECTIVE_REPEATED`) all share the same `.withSourceLocation(ext.sourceLocation)` call, so pinning a representative catches a refactor that drops the attachment from the shared call site.

The test loads SDL from a backing resource file (`src/test/resources/validation/application/scope_malformed.graphql`) because the inline-SDL overload of `fromTypeDefinitionRegistry` does not propagate a sourceName — same pattern as `NoSubscriptionsRuleTest` / `NoCustomScalarsRuleTest`.

### Local validation

- `./gradlew :core:shared:viaductschema:test` — BUILD SUCCESSFUL (full module suite green; the `ValidationContext.computeIfAbsent` addition is non-breaking for the 12 other rules' tests).
- `ScopeUsageRuleTest` — 35 tests, 0 failures (34 prior + 1 new sourceLocation regression guard).

---

_Refactor and comment generated with assistance from [Claude Code](https://claude.com/claude-code)._

xyu307:
@geovannefduarte As you can see in above comments, additional rounds of AI review and discussion ends up with 
1. More test cases even though legacy code having the right behavior but the new rules need these tests to ensure correct behavior.
2. Refactoring of parsing approach without touching test cases

I did not squash the commits into one for slice 3 so you can see the before and after of refactoring. Please see if the refactoring strike the right approach.

xyu307:
**Round 4 — re-align task input with #372 spec: `Property<SchemaScoping>` on `AssembleCentralSchemaTask`**

Following review of the slice 4 spec (#373) and a survey of slices 5–8, reverting the task input from `scopeUniverse: SetProperty<String>` to the original #372 shape `schemaScoping: Property<SchemaScoping>`. The Round 1 "minimal config-cache footprint" trade-off doesn't survive the cross-slice view:

- **Slice 5 (#374)** reads `scoping.scopedSchemas.values.toSet()` from this same task in the materialization loop. The lean input would force a follow-up PR to either add a sibling `MapProperty` input or migrate to `Property<SchemaScoping>` mid-stack.
- **Slice 8 (#377)** writes the JSON manifest from this same task and serializes both `declaredScopes` and `declaredScopedSchemas`.
- **Slice 4 (#373)** wires `includeScopeDirective = schemaScoping.get().isScoped` — a one-liner against the full shape, awkward against the universe-only input.

The original config-cache cost (data-class equality including `version: String = CURRENT_VERSION`) is real but rare-and-intentional: bumping `CURRENT_VERSION` is a breaking change to the on-disk manifest, and re-running validation at the same moment is correct, not a stray invalidation.

Diff is small (`4d0351bf`): swap the `@Input` declaration in `AssembleCentralSchemaTask` (`SetProperty<String> scopeUniverse` → `Property<SchemaScoping> schemaScoping`), change the `validScopes = ...` line in `taskAction()` to project the universe out of the data class, change the wiring line in `ViaductApplicationPlugin.setupAssembleCentralSchemaTask` to `schemaScoping.set(appExt.schemaScoping)`, and update one doc-comment in `ViaductApplicationSchemaScopeRuleTest`. Behavior is byte-identical because `schemaScoping.get().scopeUniverse` derives from the same `scopeUniverseProperty` the prior wiring read directly.

Local validation: `:gradle-plugins:application:test` green (all TestKit cases re-run against the new input shape, including the `--configuration-cache --configuration-cache-problems=fail` case); `:core:shared:viaductschema:test` unchanged.

Removes the "differs from #372 spec" note from the PR description.

core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/ValidationContext.kt
```diff
@@ -34,6 +34,27 @@ open class ValidationContext(val schema: ViaductSchema) {
...
 
+    private val ruleCaches = mutableMapOf<Pair<ValidationRule, Any>, Any?>()
+
+    /**
+     * Per-context memoization for rules that need to derive intermediate state from the schema once
+     * per (rule, key) pair. Cache lifetime equals this context, so rules need not own a long-lived
+     * cache nor worry about cross-schema reuse semantics.
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun <K : Any, V : Any> computeIfAbsent(
```
geovannefduarte, line 49: [nit] Could you add a one-line note that this assumes single-threaded validation? It's a plain `mutableMapOf` get-then-put (not getOrPut/concurrent), which is fine for the current sequential walk, but now that it's shared framework surface other rules may adopt, calling out the assumption avoids a future caller relying on atomicity.

rstata:
agent feedback:

> Blocker: framework @scope(["*"]) extensions appear to disable A.6 too broadly

more details: https://chatgpt.com/share/6a43f443-b400-83e8-af72-13c410db5248

xyu307:
## Slice 3 rebased onto the new #393 tip; review response commit added

Rebased `feat/schema-scoping-slice-3` onto the updated #393 head (`1ec158ac`, the DSL-pivot commit). Squashed the four slice-3 development commits into a single result-only commit; the intermediate refactoring rounds no longer help reviewers of the final state. A second commit responds to review.

### Rebase impact from the #393 DSL pivot

- `viaductApplication.declaredSchemaScopes(setOf(...))` → `viaductApplication { declareScoping { scopes(...) } }`. Rewrote the three `ViaductApplicationSchemaScopeRuleTest` TestKit fixtures; `ViaductApplicationScopeValidationTest` was auto-merged from #393 unchanged (slice-3 tweaks were subsumed).
- `ViaductApplicationPlugin.kt`: dropped the slice-3 `afterEvaluate { SchemaScopingValidator.validate(...) }` block (superseded by #393's hybrid-timing validation which fires at DSL-call time), and bound the extension to `val appExt` so `setupAssembleCentralSchemaTask(appExt)` can read `appExt.schemaScoping`. `AssembleCentralSchemaTask.schemaScoping: Property<SchemaScoping>` remains the wired read surface.
- All three test suites green after rebase and after the review-response commit: `ScopeUsageRuleTest` (35 → 40 cases with the new tests), `ViaductApplicationSchemaScopeRuleTest` (3 cases), `ViaductApplicationScopeValidationTest` (3 cases).

### Review responses

**@geovannefduarte** — [inline nit on `ValidationContext.kt`](https://github.com/airbnb/viaduct/pull/394#discussion_r3430627753): agreed. Added a KDoc line documenting that `computeIfAbsent(rule, key, compute)` is not thread-safe and assumes the single-threaded rule walk the framework does today. No behavior change.

**@rstata** — [Blocker: framework `@scope(["*"])` extensions appear to disable A.6 too broadly](https://github.com/airbnb/viaduct/pull/394#issuecomment-4845968349): confirmed and fixed. Root cause + fix + regression tests below.

### The blocker in one paragraph

`DefaultSchemaFactory` injects `extend type Query @scope(to: ["*"])` and `interface Node @scope(to: ["*"])`. `ParsedScopes.union` was folding every extension's raw scopes into the aggregate, including that `"*"` from the framework's star extension. `sharesScope` short-circuits `true` whenever either side contains `"*"`. Net effect: for a user-narrowed `type Query @scope(to: ["public"])`, a reference from any Query field to any scope-mismatched type silently passed A.6, because the type's effective view had been broadened to `{"public", "*"}` by an extension the user did not write. A.7 already had the inherit-base carve-out (`if (other.rawScopes == STAR_ONLY) continue`); the union view feeding A.5 / A.6 did not.

### Fix

One line in `ParsedScopes.union`, gated on `!pe.ext.isBase` so it only affects extensions:

```kotlin
for (pe in perExtension) {
    if (!pe.hasScopeDirective) continue
    if (!pe.ext.isBase && pe.rawScopes == STAR_ONLY) continue   // inherit-base carve-out
    anyDirective = true
    for (s in pe.rawScopes) {
        if (s == WILDCARD || s in validScopes) all.add(s)
    }
}
```

### Case A vs Case B — deliberate asymmetry

The carve-out is base-vs-extension aware. Under it:

| Configuration | `.union` | Behavior on `Query.f: T` where `T @scope(["internal"])` |
| --- | --- | --- |
| `type Query @scope(["public"])` + `extend type Query @scope(["*"])` (framework) | `{"public"}` | A.6 fires — reference from `["public"]` to `["internal"]` is a real mismatch. |
| `type Query @scope(["*"])` + `extend type Query @scope(["public"])` (user opts base into all scopes) | `{"*", "public"}` | A.6 does not fire — base's literal `"*"` gives universal reach. `"public" ∈ "*"` semantics. |

The asymmetry is the whole point. `"*"` on a **base** is a load-bearing user claim ("this type genuinely opts into every scope"), while `"*"` on an **extension** is the framework-injection convention meaning "extend without broadening." The extensions-only skip preserves both readings in one rule.

No A.7 change is needed for Case B: A.7's existing `val baseAllowsAll = WILDCARD in baseScopes` guard already prevents `SCOPE_EXTENSION_EXCEEDS_BASE` from firing when the base is `["*"]`. Verified by re-reading the guard — the extension's redundant narrowing on a `["*"]` base is silently accepted by A.7 today, which matches the "public is in *" reading we want.

### Test additions in `ScopeUsageRuleTest`

Five new cases pin the semantics we've discussed:

1. `A_6 framework star extension does not broaden the reference-sharing view` — rstata's exact regression scenario. Narrowed Query + framework star extension + `secret: SecretThing` where `SecretThing @scope(["internal"])` → `SCOPE_REFERENCE_NO_SHARED_SCOPE` fires on `Query.secret`.
2. `A_6 base literal star with a narrowed extension is treated as sharing every scope` — Case B pin. Base `["*"]` + extension `["public"]` → reference to a `["internal"]`-scoped type passes via the `"*"` short-circuit in `sharesScope`. Locks the "public is in *" reading and the deliberate asymmetry.
3. `A_6 star-only extension on a user type inherits the base, does not broaden` — user-written star extension on a non-framework type behaves the same as framework injection. Prevents a user smuggling `"*"` onto an extension to escape A.6.
4. `A_5 type with only a star-only extension is treated as having no @scope` — corner: base has no `@scope`, only a star-only extension. `.union` stays `null`, and a reference to the type fires `SCOPE_MISSING_ON_REFERENCED_TYPE`. Locks the semantic that the carve-out cannot silently satisfy A.5.
5. `A_6 framework Node base with narrowed Query still shares scope via wildcard` — sanity: `Node @scope(["*"])` on the **base** (not an extension) keeps its `"*"` in the union, so `Query.node: Node` from a `["public"]`-narrowed Query still shares scope. Prevents over-correction on the framework `Node` interface.

All 40 rule cases + the 3 TestKit cases pass under `--rerun`. Full command:

```
./gradlew :core:shared:viaductschema:test :core:service:api:test \
          :gradle-plugins:application:test :core:service:api:apiCheck --rerun
```

### Semantic map (for the issue thread)

I've posted a companion update on #372 with the same Case A / Case B table and the union carve-out rationale, so the spec-level record stays in sync with the code. Link: [issue #372 comment](https://github.com/airbnb/viaduct/issues/372#issuecomment-4849855405).

geovannefduarte:
I did the validation-parity check @rstata asked for.

The `@scope(["*"])` issue Raymie raised looks addressed by the latest response commit: star-only extensions are now treated as inherit-base for the aggregate sharing view, while base `["*"]` still means all scopes.

One remaining parity gap I found: the existing schema-scope validation also enforces two relationship rules that I do not see covered by `ScopeUsageRule` yet:

1. For union extensions, the extension scopes must match the intersection of the union base scopes and each added member type’s scopes.
2. For object extensions that implement an interface, the extension scopes must be within the object/interface intersection, and required interface fields must be present in those scopes.

Can you add equivalent checks and tests for those two cases?

geovannefduarte:
> I did the validation-parity check @rstata asked for.
> 
> The `@scope(["*"])` issue Raymie raised looks addressed by the latest response commit: star-only extensions are now treated as inherit-base for the aggregate sharing view, while base `["*"]` still means all scopes.
> 
> One remaining parity gap I found: the existing schema-scope validation also enforces two relationship rules that I do not see covered by `ScopeUsageRule` yet:
> 
> 1. For union extensions, the extension scopes must match the intersection of the union base scopes and each added member type’s scopes.
> 2. For object extensions that implement an interface, the extension scopes must be within the object/interface intersection, and required interface fields must be present in those scopes.
> 
> Can you add equivalent checks and tests for those two cases?

Correction on my earlier parity note: after re-checking the issue breakdown, the union/interface checks are tracked by #376 / slice 7. They are not in #372's scope, so I do not think #394 needs to add them unless we intentionally pull slice 7 work forward.

For #394, the remaining question is the behavior, including the `["*"]` base-vs-extension semantics. Could you rebase #394 on the latest head and resolve the conflicts? Once it is clean, I can run Airbnb-side validation against it before final sign-off.

xyu307:
## Rebased onto `main` — no longer stacked on slice 2

Slice 2 landed on `main` as `c846f4de` through the mirror process, so I rebased slice 3 directly onto latest `main` (`6145a783`). The branch is now a straight 2-commit sequence off `main`, no interim slice-2 tip in the middle:

```
de27d925 fix(scope): star-only extensions don't broaden the reference-sharing view
df815607 feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)
```

### Conflicts encountered and how they were resolved

Three post-slice-2 commits on `main` touched files slice 3 also modifies. Almost all of the noise traces back to one:

- **`a675d8d7 feat!: enforce schema extension correctness and add validateViaductSchemaExtensions task`** — restructured `DefaultSchemaValidator`, added `ViaductSchemaValidator(extensionsOnly=…)` mode + `SchemaExtensionsValidator`, and added `setupValidateSchemaExtensionsTask()` wiring to `ViaductApplicationPlugin`. Three-way overlap with slice-3's rule additions.
- **`b75aa2d6 feat: add build-time validation support for schema-backed @parent fields`** — added `ParentFieldConstraintsRule` to the standard rules list.
- **`647dad73 chore(oss): restore @VisibleForTest annotations`** — cosmetic, auto-merged.

**File-by-file resolutions:**

1. `core/shared/viaductschema/src/main/kotlin/viaduct/graphql/schema/validation/rules/DefaultSchemaValidator.kt`
   - `main` converted the type from `object` to `class DefaultSchemaValidator(strictMode: Boolean = false)` and expanded the rules list.
   - Kept slice-3's `standardRules: List<ValidationRule>` extraction and `create(validScopes)` opt-in overload.
   - Rebuilt `standardRules` as a `buildList { … }` so main's new entries (`ParentFieldConstraintsRule`, `PageInfoLocationRule`, the strict-mode-gated `StrictConnectionPageInfoRule`) land alongside the existing rules.
   - Net: instance-scoped rule pipeline, `strictMode` continues to work, `create(validScopes)` still appends `ScopeUsageRule` when the scope universe is non-empty.

2. `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductSchemaValidator.kt`
   - Kept main's `extensionsOnly` dispatch and chained the two changes: the non-extensions branch is now `DefaultSchemaValidator(strictMode = true).create(validScopes).validate(schema)`. When `validScopes` is `null`/empty, `create(...)` short-circuits to the same shared `validator` instance, so unscoped projects see zero behavior change.

3. `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt`
   - Kept main's `setupValidateSchemaExtensionsTask()` call and its private method definition.
   - Kept slice-3's `setupAssembleCentralSchemaTask(appExt)` signature (widened to receive the extension) and the `@OptIn(ExperimentalApi::class)` annotation, so `appExt.schemaScoping` continues to pipe into the assemble task input.

### Test verification

Full command:

```
./gradlew :core:shared:viaductschema:test :core:service:api:test \
          :gradle-plugins:application:test :core:service:api:apiCheck --rerun
```

All green:

| Suite | Tests | Failures |
| --- | ---: | ---: |
| `ScopeUsageRuleTest` | 40 | 0 |
| `DefaultSchemaValidatorTest` | 11 | 0 |
| `ViaductApplicationSchemaScopeRuleTest` | 3 | 0 |
| `ViaductApplicationScopeValidationTest` | 3 | 0 |
| `:core:service:api:apiCheck` | — | pass |

No behavior changes beyond the conflict-resolution mechanics — the review-response fixes from the previous force-push (extensions-only `STAR_ONLY` carve-out in `ParsedScopes.union`, `ValidationContext.computeIfAbsent` thread-safety KDoc, five new `ScopeUsageRuleTest` regressions) survived the rebase intact.

xyu307:
### CI green-up: rebased onto latest `main` + settings-topology fixture fix

Two rounds of upstream drift accumulated on this branch while slice-4 was under review. Both are now handled on the branch.

#### 1. `demoapps-ci-check` failures — resolved by rebase

Failing checks (all demoapps-ci-check, `build-and-test` was clean):
- `Test ktor-starter (Java 17)`, `(Kotlin 2.1.20)`
- `Test micronaut-starter (Java 21)`, `(Gradle 8.11)`, `(Gradle 9.1.0)`
- `Test jetty-starter (Java 17)`, `(Java 21)`
- `Test cli-starter (Java 17)`, `(Java 21)`

Root cause (from `Test micronaut-starter (Java 21)` job log):

```
##[error]Exception in thread "main" io.micronaut.context.exceptions.BeanInstantiationException:
  Error instantiating bean of type [viaduct.service.api.Viaduct]
1) [Guice/ErrorInCustomProvider]: SchemaScopeValidationError:
  No scope directives found from node:
  'ObjectTypeExtensionDefinition{name='Mutation', ..., fieldDefinitions=[FieldDefinition{name='echo', ...}]}' @ null:8:1
```

Slice-3's SDL-time scope enforcement was rejecting the micronaut-starter's `Mutation` extension (no `@scope`); ktor / cli / jetty jobs were cancellation-cascades from that failure.

**Fix landed on `main` after this PR's original CI run:**
- `0ec1ed95` — *fix: declare schema scopes in the ktor and micronaut starters (AIRBNB)*

Rebased branch onto latest `airbnb/main` (`9a1499a0`). Only one auto-merge in `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt`, no conflicts. Force-pushed with `--force-with-lease`.

#### 2. New `:gradle-plugins:application:test` failures — fixed in this push

Post-rebase, 9 new failures in `ViaductApplicationSchemaScopeRuleTest` (3 tests × 3 JUnit retries, across Java 11/17/21 on `ubuntu-latest` and `macos-latest`), all at the `GradleRunner.build() / buildAndFail()` lines.

Root cause: another AIRBNB commit pulled in by the rebase:
- `6aa74cad` — *feat(viaduct-gradle): Require settings topology for project plugins (AIRBNB)*

The application plugin now requires `com.airbnb.viaduct.settings-gradle-plugin` to be applied in `settings.gradle.kts` and rejects bare `rootProject.name = "..."` fixtures.

**Fix** (commit `5455e081`, this push) copies the exact pattern from the same AIRBNB commit's sibling-test update to `ViaductApplicationScopeValidationTest.kt`:

1. `writeSettings()` now applies `com.airbnb.viaduct.settings-gradle-plugin` and calls `includeViaductApplication { project(":"); modulePackagePrefix("com.example.test") }`.
2. New `combinedPluginClasspath()` helper reads `System.getProperty("java.class.path")` (needed because the settings plugin isn't on `withPluginClasspath()`'s auto-discovered classpath).
3. All three `.withPluginClasspath()` calls swapped to `.withPluginClasspath(combinedPluginClasspath())`.

No production code changed.

Local validation: `./gradlew -p gradle-plugins :application:test --tests "viaduct.gradle.ViaductApplicationSchemaScopeRuleTest"` — BUILD SUCCESSFUL (3 tests passed).

#### Slice-4 follow-up

PR #401 (slice 4) refactored this same test file to extend a new `ViaductApplicationTestKitFixture` base class. Once this PR's CI goes green, slice-4 will be rebased on top and the fixed helpers lifted into the fixture base class.

xyu307:
### Second rebase onto latest `main` — merge conflict resolved

`main` moved forward again while the previous rebase was going through CI, and the new head conflicts with slice-3.

**New AIRBNB commit on `main`:**
- `c526268e feat(viaduct): add viaduct modules bucket (AIRBNB)` — hoists a shared `viaductModules: Configuration` out of `setupAssembleCentralSchemaTask(...)` into `apply {}` and changes the function signature to take `viaductModules` as a parameter.

**Conflict** (two hunks in `gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt`):

- Call site — `main` writes `setupAssembleCentralSchemaTask(viaductModules)`; slice-3 writes `setupAssembleCentralSchemaTask(appExt)`.
- Function signature — same divergence on the sole parameter.

**Resolution** — merged both parameters. Neither intent is compromised: the function body uses `viaductModules` in `extendsFrom(viaductModules)` (from `main`) and `appExt` in `schemaScoping.set(appExt.schemaScoping)` (from slice 3).

```kotlin
val viaductModules = setupViaductModulesConfiguration()
val assembleCentralSchemaTask = setupAssembleCentralSchemaTask(viaductModules, appExt)

// ...

@OptIn(ExperimentalApi::class)
private fun Project.setupAssembleCentralSchemaTask(
    viaductModules: Configuration,
    appExt: ViaductApplicationExtension,
): TaskProvider<AssembleCentralSchemaTask> { ... }
```

**Post-rebase shape** (3 slice-3 commits on top of `c526268e`):
- `5bb49c3f feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)`
- `726eb04c fix(scope): star-only extensions don't broaden the reference-sharing view`
- `329fd55c test(scope): apply settings-topology fixture to ViaductApplicationSchemaScopeRuleTest`

**Local validation** — `./gradlew -p gradle-plugins :application:test --tests "viaduct.gradle.ViaductApplicationSchemaScopeRuleTest"` — BUILD SUCCESSFUL, 3 tests passed.

Force-pushed with `--force-with-lease`. Waiting on CI.

xyu307:
### Rebase Round 6 — onto latest `airbnb/main`

Force-pushed `feat/schema-scoping-slice-3` from `329fd55c → fbd6b2e9` (`--force-with-lease`) to pick up two upstream commits that landed on `main` since the last rebase:

- `601d8fc8` **Part 1 - Adds new Mod...** — no interaction with slice 3.
- `369c21ff` **feat(viaduct-oss): add NodeInterfaceIdConsistencyRule for schema validation** — new rule + `NODE_INTERFACE_ID_INCONSISTENT` error code, wired into `DefaultSchemaValidator`'s standard-rules list.

#### Conflicts resolved (2 files, both trivial)

1. **`core/shared/viaductschema/.../validation/ValidationErrorCodes.kt`** — both branches added a new `const` block at the same location. Kept both side-by-side (upstream's `// NodeInterfaceIdConsistencyRule` block first, then slice-3's `// ScopeUsageRule` block with its 11 `SCOPE_*` codes). Independent constants; no interaction.
2. **`core/shared/viaductschema/.../validation/rules/DefaultSchemaValidator.kt`** — upstream appended `add(NodeInterfaceIdConsistencyRule())` at the tail of `standardRules`; slice-3 had refactored that list from an inline `listOf(...)` into a hoisted `buildList { }` field consumed by both `create()` and the new `create(validScopes: Set<String>?)` overload. Kept slice-3's refactor and appended the new rule inside the `buildList`. The scoped path — `SchemaValidator(phases = listOf(standardRules + ScopeUsageRule(validScopes)))` — now composes `NodeInterfaceIdConsistencyRule` alongside `ScopeUsageRule`, which is the correct behavior (node-id consistency is a universal invariant orthogonal to scope-universe checking). The class KDoc's rule bullet list picked up the new entry via auto-merge.

#### Latent-breakage fix (amended into the third slice-3 commit)

While verifying the rebase, `:gradle-plugins:application:test` surfaced 3 pre-existing failures in `ViaductApplicationSchemaScopeRuleTest`, all with `e: Unresolved reference: modulePackagePrefix` at line 7 of the TestKit project's `build.gradle.kts`. Root cause: the test's own `buildScript(...)` helper emitted `modulePackagePrefix.set("com.example.test")` inside the `viaductApplication { }` block. That compiled on the previous base but the settings-topology work already on `airbnb/main` moved `modulePackagePrefix` out of `ViaductApplicationExtension` and made it a receiver on `includeViaductApplication { }` in `settings.gradle.kts`. Rebase surfaced the latent breakage. The peer `ViaductApplicationScopeValidationTest` avoids this because its own `buildScript` helper never emitted that line.

Fix: dropped the stale line — `writeSettings()` already sets `modulePackagePrefix("com.example.test")` inside the settings-topology block. Amended into commit `08e177c4` (title unchanged: "test(scope): apply settings-topology fixture to ViaductApplicationSchemaScopeRuleTest") → new sha `fbd6b2e9`.

#### Rebase shape (was `329fd55c`)

```
fbd6b2e9 test(scope): apply settings-topology fixture to ViaductApplicationSchemaScopeRuleTest   (was 329fd55c, now carries the latent-breakage fix)
1794c879 fix(scope): star-only extensions don't broaden the reference-sharing view              (was 726eb04c)
230fba0d feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)             (was 5bb49c3f)
369c21ff feat(viaduct-oss): add NodeInterfaceIdConsistencyRule for schema validation  ← upstream
601d8fc8 Part 1 - Adds new Mod...                                                     ← upstream
c526268e feat(viaduct): add viaduct modules bucket                                     ← prior base
```

#### Local validation (fresh M-series Apple Silicon host — Corretto 21.0.12 aarch64, Gradle 8.11, Kotlin 2.0.20)

- `./gradlew :core:shared:viaductschema:test` — **BUILD SUCCESSFUL** in 2m 56s. **89 suites, 1151 tests, 0 failures.** Includes `ScopeUsageRuleTest` (40/40) and the newly-merged `NodeInterfaceIdConsistencyRuleTest` (7/7), confirming the two rule pipelines compose cleanly on the shared `standardRules` list.
- `./gradlew :gradle-plugins:application:test` on amended head `fbd6b2e9` — **BUILD SUCCESSFUL** in 43s. **6 suites, 40 tests, 0 failures.** `ViaductApplicationSchemaScopeRuleTest` all 3 cases green (16.9s + 3.0s + 2.8s wall-clock — real `GradleRunner` TestKit invocations, real assertions on `[SCOPE_NAME_NOT_DECLARED]`, `typo`, and the configured universe in build output).

PR is now `MERGEABLE` again.

Issue #378, opened 2026-06-04
Schema Scoping: Slice 9 — Runtime reader & dual-API SchemaConfiguration factories


geovannefduarte:
This is the 9th of 12 sub-issues in the breakdown of #361. It may refer to the original consolidated [PR #359](https://github.com/airbnb/viaduct/pull/359) as an imperfect reference implementation.

**Umbrella:** #361 · **Series:** B (runtime consumption) · **Depends on:** Slice 8 · **Status:** Required

## Goal

Add the new schema-IDs-based runtime API while preserving the legacy raw-scope-set path. Implement the **dual-API model**:

> a. The runtime distinguishes the old calling flow from the new one accurately.
> b. The old calling flow has the same execution path as before.
> c. The new calling flow (schema IDs, no scope set) depends on the existence of `schema-scoping.json` or it does not work.

Plus an explicit `SchemaConfiguration.fromResources()` (no `schemaIds`) for callers who want the full schema regardless of manifest presence — this **replaces a `{FULL}` + missing-manifest escape hatch**, which had a silent-typo footgun (`"Full"` ≠ `"FULL"`).

## Files touched

- `core/service/runtime/src/main/kotlin/viaduct/service/runtime/SchemaConfiguration.kt` — new factory functions (reader logic inlined in the companion).
- `core/service/api/src/main/kotlin/viaduct/service/api/scoping/SchemaScoping.kt` — `parse(stream)` (added in Slice 8) reused here.
- Tests in `core/service/runtime/src/test/`.

### Three explicit factories

```kotlin
companion object {
    /** Full schema, no scoping. Does NOT consult the manifest. */
    fun fromResources(grtPackagePrefix: String? = null, resourcesIncluded: Regex? = null): SchemaConfiguration { ... }

    /** Scoped schemas. Reads META-INF/viaduct/schema-scoping.json, resolves each ID to its scope set, registers them.
     *  Fails-closed on missing manifest or unknown IDs. */
    fun fromResources(
        schemaIds: Set<String>,
        grtPackagePrefix: String? = null,
        resourcesIncluded: Regex? = null,
        lazyScopedSchemas: Boolean = false,
    ): SchemaConfiguration {
        require(schemaIds.isNotEmpty()) { "schemaIds must be non-empty. Use fromResources() (no schemaIds) for the full schema." }
        val scoping = readManifestOrThrow()
        val unknown = schemaIds - scoping.scopedSchemas.keys - setOf("FULL")
[...]

PR #401, opened 2026-07-14
feat(scope): gate framework @scope injection at SDL build time (slice 4 of #361)

### LINKED ISSUE #373: Schema Scoping: Slice 4 — Conditional @scope directive injection

xyu307:
**Closes #373** (slice 4 of umbrella #361).

> **Stacking:** This PR is stacked on top of **PR #394** (slice 3, branch `feat/schema-scoping-slice-3`). It needs slice 3's `Property<SchemaScoping>` task input — please review after #394 merges. Once slice 3 lands, the base retargets to `main` and only the slice-4 commit will remain in the diff.

Completes the build-time story for schema scoping: when an application does not opt into scoping via `viaductApplication.declareScoping { scopes(...) }`, `DefaultSchemaFactory` no longer emits the `@scope` directive definition, and no framework-injected type carries `@scope(to: ["*"])`. Tenant SDL that still uses `@scope` fails assembly with a Viaduct-flavored diagnostic naming two remediations — the `declareScoping` DSL, or a user-owned `directive @scope` definition. The runtime `SchemaFactory` boundary honors the same gate so `SchemaConfiguration.fromSdl/fromResources` behavior stays consistent with the Gradle task.

**What's inside:**
- `DefaultSchemaFactory.getDefaultSDL` / `addDefaults` grow an `includeScopeDirective` parameter (default `true` — every existing caller is unchanged). Five decoration sites are gated uniformly: directive definition, PageInfo, Node interface, node/nodes query extension, and the three synthetic root types. Gating is encoded on the `DefaultDirective` enum via `gatedByScoping` — no name comparisons.
- `AssembleCentralSchemaTask` and `ValidateSchemaExtensionsTask` both forward `schemaScoping.get().isScoped` to `getDefaultSDL(...)`. No new task input — `SchemaScoping.isScoped` is derived from the existing `scopeUniverse.isNotEmpty()` field bound in slice 3.
- `SchemaFactory.fromSdl(sdl, isScoped)` / `fromResources(..., isScoped)` mirror the build-time gate at the runtime boundary; `SchemaConfiguration.fromSdl(..., scopes)` and `SchemaConfiguration.fromResources(..., scopes)` companion factories derive `isScoped = scopes.isNotEmpty()`. The `SchemaFactory` defaults stay at `true` to preserve pre-slice-4 behavior for direct callers (`ResolverTestBase.getSchema()`, arbitrary-schema utilities, third-party embedders).
- `ViaductSchemaValidator.performSyntaxValidation` maps graphql-java's "Unknown directive 'scope'" / "tried to use an undeclared directive 'scope'" errors (case-insensitive) into a diagnostic that names both `viaductApplication { declareScoping { scopes(...) } }` and a user-owned `directive @scope` as valid fixes.
- Starwars opts in to `scopes("default", "extras")` — its schema uses `@scope` densely and needs the scoped path.

**Tests added:**
- 8 unit cases on `DefaultSchemaFactoryTest` covering both flag values across all five decoration sites plus SDL round-trip parsability and the public `addDefaults(registry, ...)` overload.
- 1 unit case on `ViaductSchemaValidatorTest` pinning the enriched diagnostic (both remediations mentioned).
- `ViaductApplicationScopeDirectiveGatingTest` (4 TestKit cases): scoped assembly emits the decorations; non-scoped assembly omits them; non-scoped assembly with a tenant `@scope` usage fails through the assemble task; the same failure surfaces through the preflight `validateViaductSchemaExtensions` task.
- New `ViaductApplicationTestKitFixture` abstract base class — three existing TestKit classes now share its `@TempDir`, `writeSettings`, `buildScript`, and `writeSchemaFile` helpers.

## Test plan

- [x] `./gradlew :core:shared:graphql:test` — BUILD SUCCESSFUL
- [x] `./gradlew :gradle-plugins:application:test` — BUILD SUCCESSFUL
- [x] `./gradlew :core:service:runtime:test` — BUILD SUCCESSFUL
- [x] `./gradlew :starwars:test` — BUILD SUCCESSFUL
- [x] `./gradlew :starwars:assembleViaductCentralSchema` — 7 `@scope` occurrences in `BUILTIN_SCHEMA.graphqls` (unchanged from pre-slice-4 baseline)
- [x] `./gradlew :ktor-starter:assembleViaductCentralSchema` — 0 `@scope` occurrences in `BUILTIN_SCHEMA.graphqls` (confirms the negative path)

## Open items for review

- **Regex vs. structured match.** `ViaductSchemaValidator.isUnknownScopeDirective` inspects graphql-java's rendered `message` string because both the parser-level `NonSDLDefinitionError` and the validator-level `UnknownDirectiveError` are `@Internal` in graphql-java and expose no structured "directive name" field. The `(?i)` flag handles capitalization drift; a substantive rewording would still slip past and the enrichment would silently revert to the raw error. A regression would be caught by the `Unknown scope directive parse error is enriched with both remediations` test on the next rebuild. Willing to swap for a structured check when graphql-java exposes one.
- **Diagnostic-message wording.** The enrichment naming (a) `viaductApplication { declareScoping { scopes(...) } }`, (b) a user-owned `directive @scope`, and (c) removing the `@scope` usage follows the shape of existing enrichments (e.g. the empty-output-type message). Happy to tune wording per reviewer preference — the placement, matcher, and test are the contract; the exact prose is not.
- **`@scope` on user-owned types with scoping off.** Slice 4 gates only the framework-injected `@scope` decorations. A tenant type declared with `@scope(to: [...])` when scoping is off falls into one of two paths: (i) if the tenant SDL also declares its own `directive @scope`, the build succeeds and the framework treats the directive as opaque; (ii) if the tenant SDL uses `@scope` without a definition, the build fails with the enriched diagnostic pointing at either remediation. No `@scope` usage silently no-ops.

---


xyu307:
Rebased onto the latest `feat/schema-scoping-slice-3` head (`329fd55c`) so slice-4 sits on the current slice-3 stack, which in turn now sits on the latest `airbnb/main`. New slice-4 head: `81f6b09a` (single commit replayed).

**Conflicts resolved during rebase:**

1. **`gradle-plugins/application/src/main/kotlin/viaduct/gradle/ViaductApplicationPlugin.kt`** — two hunks.
   - Call-site: `main`'s `viaductModules = setupViaductModulesConfiguration()` runs first, then the combined signature `setupAssembleCentralSchemaTask(viaductModules, appExt)` (main's `viaductModules` parameter + slice-3's `appExt` parameter), then slice-4's `setupValidateSchemaExtensionsTask(appExt)`.
   - Helpers: kept `main`'s `setupIncomingDependenciesFromTopology` and `setupViaductModulesConfiguration` verbatim, and applied slice-4's `@OptIn(ExperimentalApi::class)` + `appExt: ViaductApplicationExtension` signature to `setupValidateSchemaExtensionsTask`. Both intents preserved.

2. **`ViaductApplicationTestKitFixture.kt`** — slice-4's fixture originally wrote `settings.gradle.kts` as bare `rootProject.name = "test"` and called `.withPluginClasspath()` (no arg). Slice-3's settings-topology adaptation (chased upstream commit `6aa74cad`) requires the `com.airbnb.viaduct.settings-gradle-plugin` + `includeViaductApplication { project(":"); modulePackagePrefix(...) }` block, and a combined classpath that includes the settings plugin. Updated `writeSettings()` to emit the topology-aware script and added a protected `combinedPluginClasspath()` helper on the fixture.

3. **`ViaductApplicationSchemaScopeRuleTest.kt` and `ViaductApplicationScopeValidationTest.kt`** — slice-4's refactor onto `ViaductApplicationTestKitFixture` was accepted; the copy-pasted `@TempDir`/`writeSettings`/`buildScript`/`writeSchemaFile` blocks that had been introduced on slice-3 during its own rebase adaptation are now inherited from the fixture.

4. **`ViaductApplicationScopeDirectiveGatingTest.kt`** — no rebase conflict (new file), but its four `.withPluginClasspath()` sites needed the `combinedPluginClasspath()` argument for the same reason as slice-3's adaptation (the settings-topology plugin is not on the default `withPluginClasspath()` set for cross-included-build classpath resolution).

**Local verification:** `./gradlew :gradle-plugins:application:test` → `BUILD SUCCESSFUL in 9m 16s`, all 7 relevant TestKit suites green (45 tests, 0 failures), including the 4 new `ViaductApplicationScopeDirectiveGatingTest` cases.

Slice-4 report was updated in-repo (`worklog/slice-4-implementation-summary.md` → new "Round 2" section) with the same detail.

xyu307:
### Rebase Round 3 — restacked on slice-3 Round-6 head (`fbd6b2e9`)

Force-pushed `feat/schema-scoping-slice-4` from `81f6b09a → 94d9206f` (`--force-with-lease`) to sit on top of slice-3's Round-6 head (which itself picked up two new `airbnb/main` commits: `601d8fc8` "Part 1 - Adds new Mod..." and `369c21ff` "feat(viaduct-oss): add NodeInterfaceIdConsistencyRule for schema validation"). No slice-4 semantics changed — this rebase resolves conflicts against an upstream `SchemaFactory` refactor and drops one now-invalid line from a shared test fixture.

#### Conflicts resolved (4 files)

1. **`core/engine/wiring/src/main/kotlin/viaduct/engine/SchemaFactory.kt`** — the only load-bearing conflict. Upstream extracted a private `schemaFromTypeDefinitionRegistry(typeRegistry, coroutineInterop, customScalars, addDefaultSchemaComponents, useAppliedDirectivesOnly)` helper out of `schemaFromSdl` and added a new public `fromPrebuiltTypeDefinitionRegistry(typeRegistry)` method that reuses the helper with `addDefaultSchemaComponents = false, useAppliedDirectivesOnly = true`. Slice-4 had inlined the `DefaultSchemaFactory.addDefaults(tdr, includeScopeDirective = isScoped)` gating at the old call-site.

   Resolution: kept upstream's refactor verbatim; added `isScoped: Boolean = true` as the trailing parameter on `schemaFromTypeDefinitionRegistry`; threaded `isScoped = isScoped` from `schemaFromSdl` into the helper; moved slice-4's gating call inside the helper's existing `if (addDefaultSchemaComponents) { ... }` branch so the gate fires exactly when defaults would be added. `fromPrebuiltTypeDefinitionRegistry` intentionally has no `isScoped` parameter — its `addDefaultSchemaComponents = false` path never enters the gated branch, so the invariant "prebuilt registries are never re-decorated with `@scope`" is preserved by construction rather than by an extra guard.

2. **`gradle-plugins/application/src/test/kotlin/viaduct/gradle/ViaductApplicationSchemaScopeRuleTest.kt`** — resolved via `git checkout --theirs`. Slice-4 had already refactored this class onto `ViaductApplicationTestKitFixture`, deleting the inline `@TempDir`/`writeSettings`/`buildScript` copies. The upstream side of the conflict was Round-6's latent-breakage fix inside the now-deleted inline `buildScript` — that fix now lives structurally in the fixture (see the amended-in fix below).

3. **`gradle-plugins/application/src/test/kotlin/viaduct/gradle/ViaductApplicationScopeValidationTest.kt`** — same shape as (2), same resolution.

4. **`demoapps/starwars/build.gradle.kts`** — kept only slice-4's `declareScoping { scopes("default", "extras") }` block plus the explanatory comment; dropped the resurrected `modulePackagePrefix.set(...)` line that appeared in the three-way merge output. Verified `demoapps/starwars/settings.gradle.kts:48` still declares `modulePackagePrefix("com.example.starwars")` via `includeViaductApplication { }`, so no behavioral change.

#### Latent-breakage fix (amended into the same slice-4 commit)

`ViaductApplicationTestKitFixture` — the shared TestKit base class slice-4 introduced (Round 2) had inherited the same `appendLine("    modulePackagePrefix.set(\"com.example.test\")")` line inside its `buildScript` helper's `viaductApplication { }` block that slice-3 had to remove from its own inline `buildScript` in Round 6. Since `writeSettings()` already emits `modulePackagePrefix("com.example.test")` via `includeViaductApplication { }` in the settings-topology block, the extension-level assignment is not just redundant — it fails to compile against upstream's `ViaductApplicationExtension` (which no longer exposes `modulePackagePrefix`). Dropped the one line and amended into the same slice-4 commit (`81f6b09a → 94d9206f`, title and body unchanged), per the "amend any required change into the same commit" instruction.

#### Scope drift check against #373

Re-read issue #373 (Slice 4 — Conditional `@scope` directive injection). The rebase modifies zero lines of slice-4 semantics: no changes to `SchemaScoping.isScoped`, `DefaultSchemaFactory.getDefaultSDL(includeScopeDirective)`, `DefaultSchemaFactory.addDefaults(..., includeScopeDirective)`, `SchemaFactory.fromSdl(..., isScoped)`, `SchemaConfiguration.fromSdl/fromResources(..., scopes)`, or `ViaductApplicationPlugin` wiring. Both acceptance criteria (scoped app: framework injects the `@scope` directive definition + `@scope(to: ["*"])` decorations; non-scoped app: neither is injected and the `@scope` name is free for tenant use) are still exercised end-to-end by `ViaductApplicationScopeDirectiveGatingTest` and unit-covered across all five decoration sites by `DefaultSchemaFactoryTest`.

#### Rebase shape (was `81f6b09a`)

```
94d9206f feat(scope): gate framework @scope injection at SDL build time (slice 4 of #361)   ← this PR
fbd6b2e9 test(scope): apply settings-topology fixture to ViaductApplicationSchemaScopeRuleTest   ← slice 3
1794c879 fix(scope): star-only extensions don't broaden the reference-sharing view              ← slice 3
230fba0d feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)             ← slice 3
369c21ff feat(viaduct-oss): add NodeInterfaceIdConsistencyRule for schema validation             ← upstream
601d8fc8 Part 1 - Adds new Mod...                                                                ← upstream
c526268e feat(viaduct): add viaduct modules bucket                                                ← prior base
```

#### Local validation (fresh M-series Apple Silicon host — Corretto 21.0.12 aarch64, Gradle 8.11, Kotlin 2.0.20)

Single Gradle invocation on amended head `94d9206f`:

```
./gradlew --no-daemon :gradle-plugins:application:test :core:shared:graphql:test \
                      :core:engine:wiring:test :core:service:runtime:test
```

**BUILD SUCCESSFUL in 2m 22s. 59 suites, 582 tests, 0 failures, 0 errors, 0 skipped.**

| Module                       | Suites | Tests | Fail | Err | Skip |
|------------------------------|--------|-------|------|-----|------|
| `gradle-plugins:application` | 7      | 45    | 0    | 0   | 0    |
| `core:shared:graphql`        | 28     | 372   | 0    | 0   | 0    |
| `core:engine:wiring`         | 6      | 38    | 0    | 0   | 0    |
| `core:service:runtime`       | 18     | 127   | 0    | 0   | 0    |

The two suites most exposed to the SchemaFactory conflict resolution — `core:engine:wiring` (`SchemaFactory` runtime path) and `core:service:runtime` (`SchemaConfiguration` → `SchemaFactory.fromSdl(..., isScoped)`) — are green, confirming the gate still fires correctly after being moved into upstream's extracted `schemaFromTypeDefinitionRegistry` helper. The plugin suite covers the build-time path end-to-end via `ViaductApplicationScopeDirectiveGatingTest`'s four cases (scoped assembly + non-scoped assembly + non-scoped assemble-task failure on a tenant `@scope` usage + same failure through the preflight validate-schema-extensions task).

PR is stacked on `feat/schema-scoping-slice-3` (Round 6, `fbd6b2e9`) and now `MERGEABLE` again.

xyu307:
### Round-3 CI failure — fixed, force-pushed (`94d9206f → a90da87d`)

The Round-3 push hit two distinct failures at `airbnb/viaduct`; both amended into the same slice-4 commit (title/body/authorship unchanged), then force-pushed with `--force-with-lease=refs/heads/feat/schema-scoping-slice-4:94d9206f`. The other failing checks were matrix-cancellation cascade (`CANCELLED`, not `FAILURE`) from the first failing job — expected to recover.

#### Root cause 1 — Java interop on `DefaultSchemaFactory.addDefaults`

`JavaCodegenGoldenTest.java:215`, `JavaResolversCodegenTest.java:159`, `GraphQLSchemaParserTest.java:757` all fail:

```
error: no suitable method found for addDefaults(
    TypeDefinitionRegistry, IncludeNodeSchema, IncludeNodeSchema,
    boolean, boolean, boolean)
  method DefaultSchemaFactory.addDefaults(
    TypeDefinitionRegistry, IncludeNodeSchema, IncludeNodeSchema,
    boolean, boolean, boolean, boolean) is not applicable
```

The three Java tests pass 6 positional args; slice 4 pushed max arity to 7 (`includeScopeDirective: Boolean = true`). Java can't consume Kotlin default parameters. Fix: added `@JvmOverloads` to the public `DefaultSchemaFactory.addDefaults` method, matching the existing pattern on `ScopedSchemaBuilder.kt:63`. Auto-generated JVM overloads restore the 6-arg call site with `includeScopeDirective` defaulting to `true` (pre-slice-4 semantics for any 6-arg Java caller). No Kotlin caller affected.

Local validation:

```
./gradlew --no-daemon :core:x:javaapi:codegen:test
BUILD SUCCESSFUL in 46s
```

#### Root cause 2 — `ktor-starter` / `micronaut-starter` used `@scope` without declaring scopes

`Test micronaut-starter (Java 17)` fails at `assembleViaductCentralSchema`:

```
'Query' [@1:1] tried to use an undeclared directive 'scope' — The framework
only emits the built-in `@scope` directive definition when the application
opts into schema scoping. To fix, either (1) declare a scope universe ...
```

Both `demoapps/ktor-starter/resolvers/src/main/viaduct/schema/schema.graphqls` and `demoapps/micronaut-starter/viadapp/src/main/viaduct/schema/schema.graphqls` had `@scope(to: ["default"])` on their `extend type Query`/`Mutation` blocks, but neither demoapp's build script calls `declareScoping { scopes(...) }`. Pre-slice-4 this was a silent no-op — the directive definition was emitted unconditionally, so any usage validated. Post-slice-4 the definition is gated on `isScoped`, so tenant usages against a non-scoped app now fail assembly with exactly the diagnostic the validator was designed to emit — the CI failure is the gate working as intended, catching a pre-existing latent gap between what these demoapp SDLs claimed and what their build wiring supported.

Fix: removed the `@scope(to: ["default"])` decorations from both files. Per `demoapps/CLAUDE.md`'s explicit feature-coverage table, only `starwars` is documented to exercise "Scoped fields" — `cli-starter`, `jetty-starter`, `ktor-starter`, `micronaut-starter`, `spring-starter` are all marked absent. So removing the gratuitous decoration is the right altitude (option 3 in the diagnostic); adding `declareScoping{ }` to opt these apps into scoping would drift them off the documented coverage matrix. Verified `grep -n "declareScoping|viaductApplication|scope" demoapps/{ktor,micronaut}-starter/*/build.gradle.kts` returns nothing — no code path relies on these decorations being present.

`cli-starter`, `jetty-starter`, `spring-starter` failures were `CANCELLED` (matrix cascade), not real failures.

#### Scope drift check against #373

Neither fix touches slice-4 semantics: the `@JvmOverloads` change is annotation-only (no signature change, no Kotlin-side behavior change), and the demoapp schema edits remove SDL usage that only existed because pre-slice-4 gating was silent. Both acceptance criteria of #373 (scoped app: framework injects `@scope` definition + `@scope(to: ["*"])` decorations; non-scoped app: neither injected, `@scope` name free for tenant use) remain exercised end-to-end by `ViaductApplicationScopeDirectiveGatingTest`'s four cases plus `DefaultSchemaFactoryTest`'s 8 unit cases across all five decoration sites.

#### Rebase shape (unchanged from Round 3, tip only)

```
a90da87d feat(scope): gate framework @scope injection at SDL build time (slice 4 of #361)   ← this PR (was 94d9206f)
fbd6b2e9 test(scope): apply settings-topology fixture to ViaductApplicationSchemaScopeRuleTest   ← slice 3
1794c879 fix(scope): star-only extensions don't broaden the reference-sharing view              ← slice 3
230fba0d feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)             ← slice 3
```

PR is still stacked on `feat/schema-scoping-slice-3`.

xyu307:
### Round-4 CI failure — round-3 fix was incomplete; corrected in `a90da87d → bcd7e855`

The round-3 amend push (`a90da87d`) fixed the Java-interop compile break correctly but shipped a wrong demoapp diagnosis — I stripped `@scope(to: ["default"])` from `ktor-starter` and `micronaut-starter` on the (mistaken) reading of `demoapps/CLAUDE.md`'s coverage matrix. CI reproved the schemas are load-bearing: `micronaut-starter (Gradle 8.11)` failed at runtime with `SchemaScopeValidationError: No scope directives found from node: 'ObjectTypeExtensionDefinition{name='Mutation'...` at `SchemaScopedModule.kt:62`, and `ktor-starter` failed all six `HelloWorldTest` cases on both Java 21 and Kotlin 2.1.20 matrices. In addition, the Java-interop unblock exposed two previously-CANCELLED failures (`SimpleScopesFeatureAppTest > Unknown API deployment fails()` and 5× `PolicyCheckFeatureAppTest`) plus two ktlint violations on files this PR added.

Everything amended into the same slice-4 commit (title/body/authorship unchanged), force-pushed `a90da87d → bcd7e855` with `--force-with-lease=refs/heads/feat/schema-scoping-slice-4:a90da87d…`.

#### 1. Reverted the demoapp schema strip; opted the demoapps into scoping instead

Restored `@scope(to: ["default"])` on the Query/Mutation extensions in both `demoapps/ktor-starter/resolvers/src/main/viaduct/schema/schema.graphqls` and `demoapps/micronaut-starter/viadapp/src/main/viaduct/schema/schema.graphqls`, then added

```kotlin
viaductApplication {
    declareScoping {
        scopes("default")
    }
}
```

to `demoapps/ktor-starter/build.gradle.kts` and `demoapps/micronaut-starter/build.gradle.kts`. This is the option-2 path from the slice-4 build-time diagnostic (opt into scoping) — the right one, because these apps' runtime `SchemaScopedModule.kt:62` validator enforces that every extension in a scoped app carries a `@scope`, i.e. these apps ARE scoped, they just weren't declaring it. Pre-slice-4 the unconditional `@scope` emission masked this. Both starters now cleanly assemble + test locally with `USE_MAVEN_LOCAL=true`.

The mis-reading of `demoapps/CLAUDE.md`: its coverage table lists per-app "features exercised for the framework's benefit" (starwars gets the "Scoped fields" tick because its scope universe has multiple scopes and drives the framework's filtering tests). It is NOT a claim that the other starters can't use `@scope` at all — a starter can use `@scope(to: ["default"])` (single-scope) without needing to be a scoping-feature-coverage exemplar. Round-3 conflated those two.

#### 2. Contract-test fixture — thread scopes from `@TestSchema` SDL into `SchemaConfiguration`

`SimpleScopesFeatureAppTest > Unknown API deployment fails()` and all 5 `PolicyCheckFeatureAppTest` tests failed with `SchemaProblem at FastSchemaGenerator.java:70` — same root cause on the runtime side of the gate: their `@TestSchema` SDL contains `@scope(to: [...])` usages, but `AbstractFeatureAppTestContractBase.initViaductBuilder()` called `SchemaConfiguration.fromSdl(sdl())` with no scopes, deriving `isScoped = scopes.isNotEmpty() = false`, so the framework didn't re-emit the `directive @scope` definition and validation blew up.

Fix (in `core/tenant/runtime/src/testFixtures/kotlin/viaduct/api/testing/featureapp/AbstractFeatureAppTestContractBase.kt`): parse `@scope(to: [...])` usages out of the `@TestSchema` SDL; if any exist, register a synthetic lazy scoped-schema derivation covering the discovered scope names so `isScoped=true`. Tests that call `withScopedSchemas` still replace this configuration wholesale, so the synthetic entry only takes effect on the default path.

Slice-3 tightens `SchemaConfiguration.fromSdl` so `isScoped` is derived at the SchemaConfiguration boundary (`callers cannot accidentally desync build-time and runtime`); this fixture change respects that contract by expressing the opt-in through scopes rather than an escape hatch.

Contract-test SDL parsing uses two regexes (`@scope\(to:\s*\[…\]\)` for usages, `"([^"]*)"` for the literal list) — deliberately regex, not a full graphql-java parse, because the fixture must not itself require a scoped-schema factory to interpret the SDL it's about to hand to that same factory.

#### 3. Ktlint

`ViaductSchemaValidator.kt:205` (main sourceset) and `ViaductApplicationScopeDirectiveGatingTest.kt:23` (test sourceset) both hit `standard:function-signature` — single-expression body should fit on the function-signature line. Collapsed both. `ViaductSchemaValidator.kt:210` also flagged `standard:no-blank-line-before-body-comment` on the companion object's first commented declaration; added the blank line.

#### Local validation

```
./gradlew :core:tenant:tutorials:test --tests "…SimpleScopesFeatureAppTest"    BUILD SUCCESSFUL 37s
./gradlew :core:tenant:runtime:test --tests "…PolicyCheckFeatureAppTest"       BUILD SUCCESSFUL 39s
./gradlew :core:tenant:tutorials:test :core:tenant:runtime:test                BUILD SUCCESSFUL 1m25s
./gradlew :core:tenant:runtime:testFileBasedBootstrap :core:x:javaapi:codegen:test
                                                                               BUILD SUCCESSFUL 47s
./gradlew ktlintCheck                                                          BUILD SUCCESSFUL
./gradlew publishToMavenLocal -x test                                          BUILD SUCCESSFUL 35s
(demoapps/ktor-starter)     USE_MAVEN_LOCAL=true ./gradlew test                BUILD SUCCESSFUL 25s
(demoapps/micronaut-starter) USE_MAVEN_LOCAL=true ./gradlew test               BUILD SUCCESSFUL 23s
```

#### Scope drift check against #373

Neither section touches slice-4 semantics. Section 1 (demoapp opt-in) is the intended round-1 remediation path the build-time diagnostic emits — it recovers the pre-slice-4 behavior these apps depended on through the proper `declareScoping` DSL, matching what `starwars` already does. Section 2 (contract-test fixture) fixes a pre-existing runtime test-infrastructure gap that pre-slice-4 unconditional `@scope` emission hid; slice-4 exposed it. Section 3 is cosmetic. The four acceptance-criteria cases in `ViaductApplicationScopeDirectiveGatingTest` plus the 8 unit cases in `DefaultSchemaFactoryTest` still exercise both scoped and non-scoped assembly end-to-end.

#### Rebase shape (unchanged from round 3, tip only)

```
bcd7e855 feat(scope): gate framework @scope injection at SDL build time (slice 4 of #361)   ← this PR (was a90da87d)
fbd6b2e9 test(scope): apply settings-topology fixture to ViaductApplicationSchemaScopeRuleTest   ← slice 3
1794c879 fix(scope): star-only extensions don't broaden the reference-sharing view              ← slice 3
230fba0d feat(scope): enforce @scope universe at SDL compile time (slice 3 of #361)             ← slice 3
```

xyu307:
### CI green after single retry — Java 17 ubuntu was a Gradle-cache flake

Round-4 CI on `bcd7e855` came back mostly green with one real failure: `build-and-test / Test (Java 17) ubuntu-latest` hit a `NoSuchFileException` on `.../core/service/api/build/classes/kotlin/main/viaduct/service/api/ExecutionResult.class` during `:core:x:javaapi:runtime:compileTestJava` (javac resolved the class off the classpath, then the file was gone by the time it read the bytes). Classic Gradle build-cache / parallel-task race — a `FROM-CACHE` output getting clobbered mid-compile. The two macos reds on the same run (`Test (Java 17) macos-latest`, `Test (Java 11) macos-latest`) were secondary cancellations from the ubuntu failure aborting the run, not real failures.

Diagnosed as infra flake, not slice-4:
- Slice-4 diff (`git show --stat bcd7e855`) touches zero files in `core/service/api/` or `core/x/javaapi/runtime/`; the failing test source (`JavaVariablesProviderContractTest.java:187`) isn't modified by any commit in this PR stack.
- Java 11 ubuntu and Java 21 ubuntu both **passed** on the identical `bcd7e855`, compiling the same Java test against the same `:core:service:api` Kotlin output. Only the Java 17 matrix cell tripped.

Re-ran the failed jobs (no code change). Result:

| Job | Round-4 first run | After rerun |
|---|---|---|
| Test (Java 17) ubuntu-latest | fail 20m43s | pass 2m17s (FROM-CACHE) |
| Test (Java 17) macos-latest | cancelled 20m48s | pass 29m7s |
| Test (Java 11) macos-latest | cancelled 21m19s | pass 25m39s |
| Test (Java 21) macos-latest | (was already pending) | pass 22m10s |

All 46 checks now green. No new code push required.

core/service/runtime/src/main/kotlin/viaduct/service/runtime/SchemaConfiguration.kt
```diff
@@ -147,7 +159,10 @@ class SchemaConfiguration private constructor(
             lazyScopedSchemas: Boolean = false,
         ): SchemaConfiguration {
             return SchemaConfiguration(
-                FullSchemaConfig.FromSdl(sdl),
+                // isScoped mirrors slice-4's build-time gate: whether the framework should re-emit
+                // `directive @scope` at runtime. Non-scoped apps might carry their own `@scope`
+                // definition in tenant SDL, so we must not re-add ours.
+                FullSchemaConfig.FromSdl(sdl, isScoped = scopes.isNotEmpty()),
```
geovannefduarte: [BUG] `scopes` represents runtime schema selection, not build-time scoping opt-in. #361 explicitly allows a scoped application to register zero scoped schemas. I reproduced that case: the scoped SDL fails with `undeclared directive 'scope'` when `scopes` is empty, while the same SDL succeeds with one `ScopeConfig`.

Please carry scoping opt-in independently, or defer runtime gating until the manifest reader provides it, and add a zero-selection regression test.

core/engine/wiring/src/main/kotlin/viaduct/engine/SchemaFactory.kt
```diff
@@ -27,8 +27,20 @@ class SchemaFactory(
...
+     * @param isScoped mirrors the build-time gate on the framework's `@scope` directive definition
+     *   (see slice 4). Default is `true` — that matches the pre-slice-4 behavior of
+     *   [DefaultSchemaFactory.addDefaults] and keeps every direct [SchemaFactory] caller
+     *   (test fixtures, third-party embedders) working without change. Pass `false` when the caller
+     *   is a slice-4-aware wrapper (currently [viaduct.service.runtime.SchemaConfiguration.fromSdl])
+     *   that knows the application did not declare a scope universe — the runtime then drops
+     *   `directive @scope` and the `@scope(to: ["*"])` decorations on framework-owned types, so the
+     *   `@scope` name is free for tenant-owned use.
+     */
+    fun fromSdl(sdl: String, isScoped: Boolean = true): ViaductSchema {
```
geovannefduarte: [SUGGESTION] Kotlin default parameters do not preserve the existing JVM signatures. This removes `fromSdl(String)` and `fromResources(String, Regex)`, so existing binaries can get `NoSuchMethodError` and Java callers must pass the new Boolean.

If direct `SchemaFactory` callers remain supported, please add `@JvmOverloads` or explicit compatibility overloads

xyu307:
### Round 5 — narrowed slice-4 to build-time only; runtime `@scope` gate deferred to slice 9 (#378)

Force-pushed `bcd7e855 → 9c2ebc29` (amended into the same slice-4 commit; title/body/authorship unchanged).

Thanks for the second-round review — both of your findings were real, and the cleanest fix turned out to be your option (2): defer the runtime gate entirely to the slice that lands the manifest reader. Issue #373 was explicitly scoped as "Series: A (build-time)" and lists only `DefaultSchemaFactory`, `AssembleCentralSchemaTask`, and `DefaultSchemaFactoryTest` under "Files touched"; the runtime hops (`SchemaFactory`, `SchemaConfiguration`) were scope-creep from an earlier round on my part.

#### Reverted to slice-3 head (removes both findings by construction)

| File | Why |
|---|---|
| `core/engine/wiring/…/SchemaFactory.kt` | Dropped `isScoped` on `fromSdl` / `fromResources` / `schemaFromRuntimeSchemaFiles` / `schemaFromSdl` / `schemaFromTypeDefinitionRegistry`; `addDefaults(typeRegistry)` unconditional again. No public API change → no `@JvmOverloads` concern (finding #2 gone). |
| `core/service/runtime/…/SchemaConfiguration.kt` | Dropped the `isScoped = scopes.isNotEmpty()` derivation; `FullSchemaConfig.FromSdl` / `FromResources` no longer carry the field. No runtime opt-in signal in this slice → finding #1 gone. |
| `core/service/runtime/…/EngineRegistryTest.kt` | Reverts the slice-4-only assertions that exercised the removed runtime hop. |
| `core/tenant/runtime/…/AbstractFeatureAppTestContractBase.kt` | **Test-fixture refactor revert.** The Round-4 push added a `SCOPE_USAGE_REGEX` + `discoverScopesInSdl` + `buildInitialSchemaConfiguration()` workaround so contract-test SDL that used `@scope(to: […])` could synthesize a scoped-schema derivation to flip `isScoped=true` on. With runtime again injecting `@scope` unconditionally, that whole workaround is unnecessary. The fixture is now byte-for-byte identical to slice-3 head. |

#### Retained (the actual #373 core)

- `DefaultSchemaFactory` — `includeScopeDirective` param + five gating sites (with the earlier code-review cleanups: `DefaultDirective.gatedByScoping`, hoisted `WILDCARD_SCOPE_DIRECTIVE`, dead-defaults removed).
- `AssembleCentralSchemaTask` + `ValidateSchemaExtensionsTask` — both forward `schemaScoping.get().isScoped` so preflight and assembly gate identically.
- `ViaductSchemaValidator` — Viaduct-flavored diagnostic naming both remediations.
- `DefaultSchemaFactoryTest` (9 slice-4 cases) + `ViaductApplicationScopeDirectiveGatingTest` (5 cases including the new regression pin below).
- Demoapps `declareScoping` blocks are retained (they're needed for the *build-time* gate — tenant SDL uses `@scope`, so `AssembleCentralSchemaTask` would otherwise drop `directive @scope`). Comments on `ktor-starter` / `micronaut-starter` sharpened to name the build-time assembler explicitly.

#### New regression pin (your "add a zero-selection regression test")

`ViaductApplicationScopeDirectiveGatingTest`.`built-in schema still emits @scope when universe is declared but zero scoped schemas are registered`:

- `declareScoping { scopes("public") }` with **no** `declareScopedSchema(...)` call.
- Tenant SDL uses `@scope(to: ["public"])`.
- Asserts assembly succeeds AND the built-in SDL contains both `directive @scope` and `@scope(to: ["*"])`.
- Pins the invariant that the opt-in signal is `SchemaScoping.scopeUniverse.isNotEmpty()`, NOT the count of declared scoped schemas — so a future refactor can't silently substitute the count as a proxy.

#### Diff-size delta

| Metric | Round-4 (`bcd7e855`) | Round-5 (`9c2ebc29`) |
|---|---|---|
| Files changed vs slice-3 head | 18 | 14 |
| Net additions | 613 | 571 |
| Runtime-side files touched | 4 | 0 |

#### Local test evidence (2026-07-28)

All green:
- `:core:shared:graphql:test` — `DefaultSchemaFactoryTest` 72/0 (9 scope-specific cases).
- `:gradle-plugins:application:test` — `GatingTest` 5/0, `ValidatorTest` 12/0, `SchemaScopeRuleTest` 3/0, `ScopeValidationTest` 3/0.
- `:core:service:runtime:test` — full suite BUILD SUCCESSFUL (confirms the runtime revert is clean).
- `:core:engine:wiring:test` — full suite BUILD SUCCESSFUL (`SchemaFactory` restored signatures consistent with every caller).
- `:core:tenant:runtime:test` — full suite BUILD SUCCESSFUL (fixture revert didn't break contract-test subclasses).
- `:ktor-starter:test :micronaut-starter:test :starwars:test` — all three demoapps green.

#### Follow-up

The runtime `@scope` gate belongs in slice 9 (#378) — that's the first slice where runtime has a real `SchemaScoping` in hand (from `META-INF/viaduct/schema-scoping.json` via `readManifestOrThrow()`), which is the only correct opt-in signal. I'll leave a note on that issue naming the plumbing slice 9 needs to add and flagging the `@JvmOverloads` requirement when it re-adds an `isScoped` param to `SchemaFactory`'s public factories.

geovannefduarte (approved):

PR #403, opened 2026-07-29
feat(scope): materialize and validate scoped schemas at build time (slice 5 of #361)

### LINKED ISSUE #374: Schema Scoping: Slice 5 — Scoped-schema materialization & introspection

xyu307:
Closes #374 (slice 5 of umbrella #361).

Stacked on top of #401 (slice 4). The diff here is cumulative against `main` — slice-5-only changes are the top commit `feat(scope): materialize and validate scoped schemas at build time`. Review by comparing against `feat/schema-scoping-slice-4`, or wait until #401 lands.

## Summary
- Wires `ScopedSchemaValidator` into `AssembleCentralSchemaTask`.
- For every scoped application, enumerates distinct scope sets — declared `scopedSchemas` values plus the always-added universe set — materializes each filtered projection via `ScopedSchemaBuilder`, and asserts introspectability with `IntrospectionQuery.INTROSPECTION_QUERY`.
- Failures are aggregated across all sets before the task throws, so operators get one error report per build rather than one per Gradle re-run.
- Scope sets iterate in canonical sorted order (`sortedBy { it.toSortedSet().toString() }`) so INFO/WARN/ERROR log ordering is stable across runs.
- New error code: `ValidationErrorCodes.SCOPED_SCHEMA_INTROSPECTION_FAILED`.

## D3 deviation
Empty-set aliases (`scopedSchema("FULL_ALIAS")` with no scope IDs) collapse into the always-added universe set instead of being skipped per the spec's literal `if (scopeSet.isEmpty()) continue`. This gives well-defined "materialize as base" semantics and guarantees scoped apps with zero declared `scopedSchema(...)` calls still get build-time coverage of the `SchemaId.Base` projection.

## Cross-module dependency note
`core/shared/graphql` cannot depend on `core/shared/viaductschema` (reverse edge already exists). `ScopedSchemaValidator.Failure` is a helper-local data class; the caller (`AssembleCentralSchemaTask`) reads `ValidationErrorCodes.SCOPED_SCHEMA_INTROSPECTION_FAILED` on its own side of the module boundary and formats the diagnostic there.

## Follow-up (deferred, not blocking this PR)
Assembled schema files are currently parsed twice — once inside `validateCompleteSchema` (via `ViaductSchemaValidator`), then again in `parseSchemaForScopedValidation`. Fixing this requires threading the parsed `TypeDefinitionRegistry` (or `GraphQLSchema`) out of `ViaductSchemaValidator`, which is a non-trivial refactor of the existing validator's return surface with its own review scope. Deferring to a separate PR to keep slice-5 review focused.

## Self-review pass
A `/code-review` pass over the slice-5 diff surfaced 7 findings; 5 were fixed here, 1 was kept-with-doc-fix (`validatedScopeSets` — test-only observability state, deleting it would drop three unit tests of contract coverage), and the double-parsing issue above is the deferred one.

## Test plan
- [x] `./gradlew :core:shared:graphql:test --tests ScopedSchemaValidatorTest` — 8/8 pass
- [x] `./gradlew :gradle-plugins:application:test --tests ViaductApplicationScopedSchemaMaterializationTest` — 5/5 pass
- [x] `./gradlew :core:shared:graphql:ktlintCheck :core:shared:graphql:detekt :core:shared:viaductschema:ktlintCheck :core:shared:viaductschema:detekt :gradle-plugins:application:ktlintCheck :gradle-plugins:application:detekt` — BUILD SUCCESSFUL

## Test coverage
- `ScopedSchemaValidatorTest` (unit): success on single/multiple scope sets, `validatedScopeSets` accumulation contract, failure surfacing when root type is filtered away, failure-message pins the scope set (never the alias), no-op / empty-input safety, universe-set (BASE-equivalent) materialization with `@scope(to: ["*"])`.
- `ViaductApplicationScopedSchemaMaterializationTest` (TestKit): empty-set alias resolves to universe (D3 pin), failure diagnostic names scope set not alias, zero-declared-scoped-schemas still materializes universe, unscoped app skips materialization entirely, per-scope-set INFO timing lines under `--info` + scope-set dedup.


core/shared/graphql/src/main/kotlin/viaduct/graphql/scopes/ScopedSchemaValidator.kt
```diff
@@ -0,0 +1,106 @@
...
+     * @return empty list on success; one or more [Failure]s on introspection or materialization
+     *         error. Never throws — materialization exceptions are captured as failures so the
+     *         caller can aggregate across all scope sets before reporting.
+     */
+    fun validate(scopeSet: Set<String>): List<Failure> {
+        _validatedScopeSets.add(scopeSet)
+
+        val isUniverse = scopeSet == validScopes.toSet()
+        val filtered: GraphQLSchema = try {
+            builder.applyScopes(scopeSet, includeTenantLocalFields = isUniverse).filtered
```
geovannefduarte: For the universe set, `includeTenantLocalFields = true` selects the transformer branch that skips filtering, type removal, and `ValidateScopesVisitor`. Using the existing directive-retention fixture, this validator succeeds while normal `applyScopes(validScopes)` throws `DirectiveRetainedTypeScopeError`. This leaves A.8 unchecked when no explicit subset exists and introspects a different shape from runtime Base. Could we materialize the actual projection and run required-scope validation separately?

core/shared/graphql/src/main/kotlin/viaduct/graphql/scopes/ScopedSchemaValidator.kt
```diff
@@ -0,0 +1,106 @@
...
+        } catch (e: Exception) {
```
geovannefduarte, line 90: `SchemaScopeValidationError` extends `Throwable`, not `Exception`, so this catch misses it and subclasses such as `DirectiveRetainedTypeScopeError`. The exception escapes `validate()`, aborts validation of remaining scope sets, and bypasses the aggregated `SCOPED_SCHEMA_INTROSPECTION_FAILED` diagnostic. Please catch `SchemaScopeValidationError` explicitly and add an aggregation regression test.

core/shared/graphql/src/main/kotlin/viaduct/graphql/scopes/ScopedSchemaValidator.kt
```diff
@@ -0,0 +1,106 @@
...
+import graphql.GraphQL
+import graphql.introspection.IntrospectionQuery
+import graphql.schema.GraphQLSchema
+import java.util.SortedSet
+
+/**
+ * Build-time helper that materializes filtered schema projections for each declared scope set and
+ * asserts that each projection is introspectable.
+ *
+ * This is the load-bearing helper for slice 5 of the schema-scoping DSL (#361 / #374). The caller
```
geovannefduarte: This public KDoc mixes the API contract with rollout history and current repository topology: “slice 5,” issue numbers, a test method name, and module dependency direction. It also states that runtime `SchemaId.Base` keeps tenant-local fields, while `EngineRegistry` builds Base with `applyBaseSchema()`, whose tests verify those fields are removed. Please reduce this to durable inputs, outputs, and invariants; keep slice/review history in the PR or issue

core/shared/graphql/src/main/kotlin/viaduct/graphql/scopes/ScopedSchemaValidator.kt
xyu307: Addressed in 888271fa. The `includeTenantLocalFields = isUniverse` toggle is gone — `validate()` now unconditionally passes `false`, matching runtime `SchemaId.Scoped`. To materialize the actual BASE projection (which is `applyBaseSchema()` post-`0e074c46`, not `applyScopes(universe, includeTenantLocalFields = true)`), a separate `validateBase()` entry point calls `builder.applyBaseSchema().filtered` and runs the same introspection check.

`AssembleCentralSchemaTask.materializeAndIntrospectProjections` (renamed from `validateScopedSchemas`) drives both: `validateBase()` runs for every application (scoped and unscoped) outside the `isScoped` gate, then the declared-scope-set loop runs inside the gate with the tenant-local flag hardwired to `false`. A.8 (`DirectiveRetainedTypeScopeError` from the directive-retention fixture) now fires against BASE via `applyBaseSchema()`, matching the runtime shape.

Broader plan reconciliation, including why the universe-with-tenant-local-IN shape has no runtime counterpart post-`0e074c46`, is in [airbnb/viaduct#374 (comment)](https://github.com/airbnb/viaduct/issues/374#issuecomment-5122666917).

core/shared/graphql/src/main/kotlin/viaduct/graphql/scopes/ScopedSchemaValidator.kt
xyu307: Fixed in 888271fa. Both `validate()` and `validateBase()` now catch `SchemaScopeValidationError` explicitly ahead of the generic `Exception` branch, so `DirectiveRetainedTypeScopeError` (and any other Throwable-based `SchemaScopeValidationError` subclass) is converted into a `Failure` record instead of escaping.

Aggregation regression test: `ScopedSchemaValidatorTest."sequential validate calls after a SchemaScopeValidationError all run and record"` — asserts that after `validate(setOf("other-scope"))` returns a failure (rather than propagating), a subsequent `validate(setOf("test-scope"))` still runs and is recorded in `validatedScopeSets`. Before the fix, the first call's Throwable aborted the loop and the second scope set was silently skipped.

core/shared/graphql/src/main/kotlin/viaduct/graphql/scopes/ScopedSchemaValidator.kt
xyu307: Rewritten in 888271fa. The KDoc now describes only durable inputs, outputs, and invariants of the two entry points (`validate`, `validateBase`), plus why the API is scope-set-keyed rather than alias-keyed. Slice/issue references and the test-method name are gone. The load-bearing module-topology note is kept as a brief sentence explaining why `Failure` is a helper-local data class rather than a `SchemaValidationError` — happy to move that out entirely if you'd prefer it live in a package `AGENTS.md` or the class implementation instead of the KDoc.

The factually wrong claim about `SchemaId.Base` retaining tenant-local fields is corrected: `validateBase()` now materializes via `applyBaseSchema()`, matching what `EngineRegistry` builds at runtime, and the KDoc reflects that.

xyu307:
## Follow-up push: 888271fa

Additive commit on top of the slice-5 head (no amend of the original squash target) so this delta is reviewable standalone. Rationale for the shape of these changes is in the updated [airbnb/viaduct#374 (comment)](https://github.com/airbnb/viaduct/issues/374#issuecomment-5122666917) — five-point reconciliation with the `FULL` → `BASE` change on main (commit `0e074c46`).

### What changed

**Runtime-shape alignment (`ScopedSchemaValidator`, `core/shared/graphql`)**
- New `validateBase()` materializes via `ScopedSchemaBuilder.applyBaseSchema().filtered` — matching runtime `SchemaId.Base` exactly. Was previously approximated as `applyScopes(universe, includeTenantLocalFields = true)`, which post-`0e074c46` is a different code path (no `FilterTenantLocalFieldsVisitor`, no `ValidateScopesVisitor`).
- Dropped the `includeTenantLocalFields = isUniverse` toggle from `validate()`; always `false`, matching runtime `SchemaId.Scoped`.
- Explicit `catch (SchemaScopeValidationError)` branch — it extends `Throwable`, not `Exception`, so the previous `catch (Exception)` let `DirectiveRetainedTypeScopeError` (and friends) escape and abort aggregation across scope sets. Regression test pins the contract.
- Rewrote the KDoc to describe durable inputs/outputs/invariants only; dropped slice/issue/test-method references and the wrong tenant-local claim.

**Task wiring (`AssembleCentralSchemaTask`, `gradle-plugins/application`)**
- Renamed `validateScopedSchemas` → `materializeAndIntrospectProjections` — the method fires an entirely separate rule surface (scope-visitor validators via `ScopedSchemaBuilder.apply*`) plus an introspection query on each projection, in addition to the checks `validateCompleteSchema` already runs on the raw SDL.
- Called unconditionally; BASE materialization moved outside the `if (scoping.isScoped)` gate because BASE is the runtime default for both scoped and unscoped applications. Per-scope-set loop stays inside the gate.
- Aggregated `(label, message)` failures with a `:BASE` sentinel (leading colon guarantees no collision with scope-set labels, which start with `[`). Error output distinguishes "BASE schema" vs. "scope set [foo, bar]"; summary counts them separately.

### Review threads addressed

- [Comment on `ScopedSchemaValidator.kt:75`](https://github.com/airbnb/viaduct/pull/403#discussion_r3675614731) — `includeTenantLocalFields = isUniverse` and A.8 coverage — replied inline.
- [Comment on `ScopedSchemaValidator.kt:90`](https://github.com/airbnb/viaduct/pull/403#discussion_r3675626253) — `SchemaScopeValidationError` extends `Throwable` — replied inline.
- [Comment on `ScopedSchemaValidator.kt:12`](https://github.com/airbnb/viaduct/pull/403#discussion_r3675645320) — KDoc mixes API contract with rollout history — replied inline.

### Verification

- `:core:shared:graphql:test` — `ScopedSchemaValidatorTest` 11/11 pass, including three new cases: `validateBase` clean-schema success, `SchemaScopeValidationError` captured not propagated, and sequential-validate aggregation-after-error contract.
- `:gradle-plugins:application:test` — `ViaductApplicationScopedSchemaMaterializationTest` 5/5 pass; scenarios 5/7/8/9 updated to assert the two-lane BASE + per-scope-set contract.

`mergeable_state: dirty` on this PR is a merge conflict against `main` from the other in-flight umbrella slices, not from this commit — will rebase once the queue clears or on request.

PR #414, opened 2026-08-20
ci(alerts): keep the coverage summary advisory and label re-run alerts

geovannefduarte:
## Description

**A transient dependency failure turns `main` red through an advisory job.** In run [32311746606](https://github.com/airbnb/viaduct/actions/runs/32311746606), `Coverage Summary` failed with `Artifact not found for name: coverage-core`. `Coverage (core)` could not resolve `org.jacoco:org.jacoco.ant:0.8.12`; Maven Central returned `429`. That step's `|| { echo "::warning::..." }` wrapper reported the job as successful. No reports were produced, so no artifact was uploaded. `coverage-summary`'s `needs.coverage-reports.result == 'success'` gate then passed, and `download-artifact` failed the job. The gate works as written. The defect is that `coverage-summary` requires artifacts an advisory job may not produce.

**Re-run alerts read identically to the original.** Run 32311746606 is attempt 2 and alerted twice with the same text. Runs [32199780064](https://github.com/airbnb/viaduct/actions/runs/32199780064) and [32285827987](https://github.com/airbnb/viaduct/actions/runs/32285827987) alerted on attempt-1 failures that passed on re-run.

### Changed

`coverage-summary` gets `continue-on-error: true` on its three `download-artifact` steps, and a
`!cancelled()` gate so one failed project no longer discards the summary for the other two.
`format_coverage_summary.py` already prints `⚠️ No per-module coverage reports found.` when it finds
none, so it needs no change.

`format_alert.py` gains an optional `attempt`, rendered only above 1:

```
before  :red_circle: Build and Test failed on `main` — commit `4ca5b19` by viaductbot (https://…)
after   :red_circle: Build and Test failed on `main` — commit `4ca5b19` by viaductbot, attempt 2 (https://…)
```

`.github/actions/collect-failure-info` already wrapped `format_alert.py` for four workflows.
`ci-trigger.yml` duplicated its payload construction inline, and now uses the action. All five call
sites pass `attempt`. `ci-watchdog.yml` passes `github.event.workflow_run.run_attempt`, because its
alert describes the observed run rather than the watchdog.

- `actionlint` across every workflow.
- `python3 -m unittest tests.test_format_alert tests.test_format_coverage_summary`.
- The composite action's `jq` payload piped to `format_alert.py` using run 32311746606's values at attempt `""`, `1` and `2`, and with a two-job list. Exit 0 each time, output as quoted above.

### DIFF
--- a/.github/actions/collect-failure-info/action.yml
+++ b/.github/actions/collect-failure-info/action.yml
@@ -22,6 +22,10 @@ inputs:
     description: 'GitHub actor (optional)'
     required: false
     default: ''
+  attempt:
+    description: 'Run attempt number (optional); labeled in the alert only when above 1'
+    required: false
+    default: ''
 
 outputs:
   text:
@@ -41,8 +45,9 @@ runs:
         REPOSITORY: ${{ inputs.repository }}
         SHA: ${{ inputs.sha }}
         ACTOR: ${{ inputs.actor }}
+        ATTEMPT: ${{ inputs.attempt }}
       run: |
-        # Build the full payload — format_alert.py handles empty sha/actor
+        # Build the full payload — format_alert.py handles empty sha/actor/attempt
         # gracefully via .get() + truthiness checks.
         text=$(jq -n \
           --arg branch "$BRANCH" \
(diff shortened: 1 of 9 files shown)


geovannefduarte:
...

